PR #1, opened 2015-12-01
Change to using libyaml-dev, to speed up pyyaml

hashbrowncipher:
If we have to build our own wheel, we need libyaml-dev headers in order
to get C accelerated yaml parsing.


solarkennedy:
This is in master now.

PR #3, opened 2015-12-23
Split startup out into its own script; simplify reaping

hashbrowncipher:
While working on SMTSTK-155 I realized that reaping could also be made significantly simpler.  Attached is my first draft.


bobtfish:
lgtm

PR #4, opened 2016-01-07
Use c yaml bindings

jolynch:
The native python yaml parser is _really_ slow, so we should build
against the libyaml-dev package so that we can take advantage of c
bindings if we have them

### DIFF
--- a/dockerfiles/itest/itest/Dockerfile.lucid
+++ b/dockerfiles/itest/itest/Dockerfile.lucid
@@ -24,9 +24,10 @@ ADD habitat /nail/etc/habitat
 ADD ecosystem /nail/etc/ecosystem
 ADD region /nail/etc/region
 ADD itest.py /itest.py
+ADD run_itest.sh /run_itest.sh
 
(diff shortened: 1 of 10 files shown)


EvanKrall:
👍 looks good to me, pending trusty build

asottile:
This is insufficient, you'll also need to use the CLoader in your load() call

Here's the code I tend to use:

```python
try:
    from yaml import CLoader as Loader
except ImportError:
    from yaml import Loader

yaml.load(..., Loader=Loader)
```

EvanKrall:
If we're going to dep libyaml, we might as well just do
```
from yaml import CLoader

yaml.load(..., Loader=CLoader)
```

jolynch:
@EvanKrall, @asottile The yaml loading that this package does actually doesn't matter, the issue is service_configuration_lib which [does](https://github.com/Yelp/service_configuration_lib/blob/9577c493c3c62eeaceda5d35c131bf3a35a380e7/service_configuration_lib/__init__.py#L58-L62) use the proper loader.

PR #7, opened 2016-04-22
When adding extra headers, delete those headers.

bobtfish:
Internal: https://reviewboard.yelpcorp.com/r/145730/


vulpine:
Ship it!

PR #11, opened 2016-11-15
added backup backends to synapse

mjksmith:
Added functionality to configure_synapse to create multiple backends for a service (one for each advertise type) and like them to one frontend using  haproxy ACLs.

@solarkennedy @jolynch @jnb

### DIFF
--- a/dockerfiles/itest/itest/itest.py
+++ b/dockerfiles/itest/itest/itest.py
@@ -27,7 +27,7 @@ SERVICES = {
         'mode': 'http',
         'healthcheck_uri': '/my_healthcheck_endpoint',
         'discover': 'habitat',
-        'advertise': ['habitat', 'advertise'],
+        'advertise': ['habitat', 'region'],
     },
 
     # TCP service
--- a/dockerfiles/itest/itest/itest.py
+++ b/dockerfiles/itest/itest/itest.py
@@ -93,7 +93,7 @@ def setup():
         # Fake out a nerve registration in Zookeeper for each service
         for name, data in SERVICES.iteritems():
             zk.create(
-                path=('/nerve/all/%s/itesthost' % name),
+                path=('/smartstack/global/%s/itesthost' % name),
                 value=(json.dumps({
                     'host': data['ip_address'],
                     'port': data['port'],
@@ -146,9 +146,13 @@ def test_synapse_services(setup):
     expected_services = [
         'service_three.main',
         'service_three.main.region',
+        'service_three.main.remote',
         'service_one.main',
+        'service_one.main.remote',
         'service_three_chaos.main',
+        'service_three_chaos.main.remote',
         'service_two.main',
+        'service_two.main.remote',
     ]
 
     with open('/etc/synapse/synapse.conf.json') as fd:
@@ -165,7 +169,7 @@ def test_http_synapse_service_config(setup):
         'discovery': {
             'hosts': [ZOOKEEPER_CONNECT_STRING],
             'method': 'zookeeper',
-            'path': '/nerve/all/service_three.main',
+            'path': '/smartstack/global/service_three.main',
             'label_filters': [
                 {
                     'label': 'habitat',
@@ -194,6 +198,8 @@ def test_http_synapse_service_config(setup):
                 'use_backend service_three.main if service_three.main_has_connslots',
                 'acl service_three.main.region_has_connslots connslots(service_three.main.region) gt 0',
                 'use_backend service_three.main.region if service_three.main.region_has_connslots',
+                'acl service_three.main.remote_has_connslots connslots(service_three.main.remote) gt 0',
+                'use_backend service_three.main.remote if service_three.main.remote_has_connslots',
                 'default_backend service_three.main',
             ],
             'backend': [
@@ -217,7 +223,7 @@ def test_backup_http_synapse_service_config(setup):
         'discovery': {
             'hosts': [ZOOKEEPER_CONNECT_STRING],
             'method': 'zookeeper',
-            'path': '/nerve/all/service_three.main',
+            'path': '/smartstack/global/service_three.main',
             'label_filters': [
                 {
                     'label': 'region',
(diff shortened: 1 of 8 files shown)


mjksmith:
@jolynch and I talked about this and it needs a couple changes from the nerve end and it's handling of `extra_advertise` before it's ready.

mjksmith:
Right now this generates remote backends even for services that don't have `extra_advertise`. If `extra_advertise` isn't set they'll do nothing but it might worth be doing some introspection to only create remote backends if `extra_advertise` is present and applicable to reduce confusion (why are there empty remote backends everywhere??? etc.)

Also I'm using `connslots($BACKEND) gt 0` to determine if a request should be proxied to the next backend which isn't the safest (there could still be a race condition). I'm also not too sure about the performance of `connslots`. We could use a different method to determine if a service is safe (haproxy guides recommend checking to see if the number of connections exceeds a set limit equal to some percentage of the actual max connections but that was written before `connslots` existed). We could also change `0` to a percentage of the max connection count (e.g. 5%).

If anyone could weight in about these that would be appreciated.

mjksmith:
@jolynch problem with the current `remote` backend method:

Say we have three habitats, `h1`, `h2` and `h3` and a `smartstack.yaml` that looks like:
```
---
main:
  discover: habitat
  advertise: ['habitat']
  extra_advertise:
    habitat:h1:
    - habitat:h2
  ...  etc  ...
```

Synapse in `h3` will still create a `remote` backend that can access services running in `h1` since the backend will look for any services that matches the condition `habitat != h3` which will match the `h1` services `extra_advertise`-ed to `h2`.

This can be fixed for this case by not creating `remote` backends in areas not affected by `extra_advertise` but still fails in more complicated cases. e.g.:
```
---
main:
  discover: habitat
  advertise: ['habitat']
  extra_advertise:
    habitat:h1:
    - habitat:h2
    - habitat:h3
  ...  etc  ...
```

Not sure what we can do with `extra_advertise` to fix this.

src/synapse_tools/configure_synapse.py
```diff
@@ -155,26 +157,125 @@ def generate_base_config(synapse_tools_config):
...
+            # don't create acls that downcast requests
+            continue
+
+        backend_identifier = get_backend_name(
+            service_name=service_name,
+            discover_type=discover_type,
+            advertise_type=advertise_type,
+        )
+
+        # use connslots acl condition
```
solarkennedy: This needs some docs. I'm trying to figure out how `connslots` work, and a hint would be nice here.

It also isn't super obvious, but can you confirm this is the only ACL you need? (connslots > 0 and the default_backend) I'm surprised it is that simple, but cool!

src/tests/configure_synapse_test.py
```diff
@@ -80,15 +111,112 @@ def test_generate_configuration(mock_get_current_location):
                     'capture request header X-B3-Flags len 10',
                     'capture request header X-B3-Sampled len 10',
                     'option httplog',
+                    'acl test_service_has_connslots connslots(test_service) gt 0',
```
mjksmith: We don't use dynamic connections AFAICT and I don't think we set `maxconn` or `maxqueue` to 0 but if we do it looks like this will use the correct behaviour (the maximum connection count is 0 so we should move on to the next backend).

dockerfiles/itest/itest/itest.py
```diff
@@ -133,7 +144,17 @@ def test_synapse_qdisc_tool(setup):
 
 
 def test_synapse_services(setup):
-    expected_services = ['service_three.main', 'service_one.main', 'service_three_chaos.main', 'service_two.main']
+    expected_services = [
+        'service_three.main',
+        'service_three.main.region',
```
jolynch: I'm not sure we want a test of crossing zks, but let's have it in our test plan for dev.

src/synapse_tools/configure_synapse.py
```diff
@@ -155,26 +158,124 @@ def generate_base_config(synapse_tools_config):
...
+        frontend_acl_configs.extend(
+            [
+                'acl {backend_identifier}_has_connslots connslots({backend_identifier}) gt 0'.format(
```
jolynch: We don't set maxconn on our server instances, which I believe means that we get the default of 0 ([unlimited](https://cbonte.github.io/haproxy-dconv/1.5/configuration.html#5.2-maxconn)) I think we need either:

1. Set maxconn on frontends (I sorta like this)
2. Set maxconn on servers (I sorta like this too, set it to like 100 or something)
3. Us nbsrv acls instead

src/synapse_tools/configure_synapse.py
```diff
@@ -155,26 +158,124 @@ def generate_base_config(synapse_tools_config):
...
+        advertise_types = sorted(
+            service_info.get('advertise', ['region']),
+            key=lambda typ: location_depth_mapping[typ],
+            reverse=True,  # consider the most specific types first
         )
+        if discover_type not in advertise_types:
+            return {}
 
-    return synapse_config
+        proxy_port = service_info['proxy_port']
```
jolynch: nit: I'd move this up to where we do the None check on proxy_port (so if someone changes that they know to change this). Actually I'd just combine the two.

```
proxy_port = service_info.get('proxy_port')
if proxy_port is None:
    continue
```

src/synapse_tools/configure_synapse.py
```diff
@@ -155,26 +158,124 @@ def generate_base_config(synapse_tools_config):
...
+
+            config['discovery']['label_filters'] = [
+                {
+                    'label': advertise_type,
+                    'value': get_current_location(advertise_type),
+                    'condition': 'equals',
+                },
+            ]
+
+            synapse_config['services'][backend_identifier] = config
```
jolynch: Let's explicitly set backend_name so we guarantee we can route with ACLs

https://github.com/airbnb/synapse/#the-haproxy-section

src/synapse_tools/configure_synapse.py
```diff
@@ -155,26 +158,124 @@ def generate_base_config(synapse_tools_config):
...
+
+        # create the remote backend
+        remote_config = copy.deepcopy(base_haproxy_cfg)
+        remote_config['discovery']['label_filters'] = [
+            {
+                'label': 'remote_%s' % discover_type,
```
jolynch: I think this may warrant a comment on what nerve is doing here (e.g. "Nerve creates remote entries for every instance of the discovery type, so Synapse just has to read the 'local' one" or something ...

dockerfiles/itest/itest/yelpsoa-configs/service_two/smartstack.yaml
```diff
@@ -7,6 +7,8 @@ main:
   healthcheck_uri: /lil_brudder
   allredisp: False
   discover: habitat
+  advertise:
```
mjksmith: It should just default to `advertise: ['region']`

It's probably better for a test instead of an itest so I'll add it there

src/synapse_tools/configure_synapse.py
mjksmith: Default is 2000 AFAICT

https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4.2-maxconn

mjksmith:
@jolynch I implemented the changes you suggested. Can you rereview and approve changes if all looks good?

jolynch:
👍 👍 👍 👍

PR #13, opened 2016-12-14
Have haproxy-synapse expose unix sockets for each service

jglukasik:
Done in `configure_synapse.py` by injecting a new `bind /var/run/synapse/sockets/%s.sock` line into haproxy.conf.json, to be read by synapse in setting up its haproxy instance.

The tests were updated to reflect this new entry in the 'frontend' section of the generated json. A small helper function was added in the itest to sort lists nested inside dicts. When testing this, I didn't craft the mock json in the same way it was being generated, causing failures due to the differences in list orderings.

Installing this new synapse-tools package will cause /var/run/synapse/sockets/ to be populated with a unix socket for each service, as long as that directory is already created (we'll have to do this in puppet).

Using our internal `proxycurl` based on a recent version of curl, testing these sockets with `sudo proxycurl --unix-socket /var/run/synapse/sockets/devops.demo.sock http://localhost/status` returns the status json, just like curl'ing the TCP/IP endpoint.

### DIFF
--- a/dockerfiles/itest/itest/itest.py
+++ b/dockerfiles/itest/itest/itest.py
@@ -184,6 +184,7 @@ def test_http_synapse_service_config(setup):
                 'timeout server 11000ms'
             ],
             'frontend': [
+                'bind /var/run/synapse/sockets/service_three.main.sock',
                 'timeout client 11000ms',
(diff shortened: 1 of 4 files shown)


solarkennedy:
lgtm

PR #17, opened 2017-01-12
give unix sockets permissions 777

jglukasik:


src/synapse_tools/configure_synapse.py
```diff
@@ -83,7 +83,9 @@ def generate_base_config(synapse_tools_config):
 
                 # Send syslog output to syslog2scribe
                 'log 127.0.0.1:1514 daemon info',
-                'log-send-hostname'
+                'log-send-hostname',
+                'unix-bind mode 777'
```
solarkennedy: I think 666 is actually the 'best' permission we can give this for our use case.

fix and ship

PR #19, opened 2017-01-25
Added in proxied_through to configure_synapse

avadhutp:
Testing pending. Ticket for reference: https://jira.yelpcorp.com/browse/PERF-1196.


avadhutp:
cc @ArnaudBrousseau @drolando @solarkennedy

drolando:
Do we have docs for this project? Shouldn't we write somewhere what fronted_by is and when to use it?

avadhutp:
@drolando I think that documentation will go into paasta docs, no? For now, I am following whatever's in the [CEP](https://docs.google.com/document/d/1r3AP6PS7UV5iKun6_GQFJB_5W4H8mi28VgTF6aLNIuY/edit#heading=h.tzzp8p5sn72z).

drolando:
My point is just that this is a public repo, so we should have at least a minimal documentation of what each option can be used for. I was thinking of 2 lines, not a full doc page

avadhutp:
@drolando You mean in the code itself? I did not see any in-code doc for, lets say, `advertise`. The README in this repo does have a link to the actual PaaSTA documentation: http://paasta.readthedocs.io/en/latest/yelpsoa_configs.html#smartstack-yaml.

src/synapse_tools/configure_synapse.py
```diff
@@ -182,6 +182,22 @@ def generate_acls_for_service(service_name, discover_type, advertise_types):
             advertise_type=advertise_type,
         )
 
+        # check for fronted_by first, use_backend ordering matters
+        if fronted_by:
+            frontend_acl_configs.extend(
+                [
+                    'acl request_from_proxy hdr_beg(X-Yelp-Source) -i {fronted_by}'.format(
```
solarkennedy: How about `X-SmartStack-Fronted_by`?

solarkennedy:
I think you need to upgrade the `six` package to get the build green

src/synapse_tools/configure_synapse.py
avadhutp: Done.

avadhutp:
@solarkennedy I don't think `six` is a problem. This happens for all <xenial builds. `setuptools` > 34 requires six causing tox to fail. See https://github.com/pypa/setuptools/issues/938.

mattmb:
Yeah this has been causing me pain the last couple of days.
https://github.com/Yelp/paasta/pull/976 was the fix in paasta.

avadhutp:
Fixed trusty builds, thanks to @mattmb. However, lucid is still problematic. I was able to fix the `six` dependency doing [this](https://github.com/Yelp/synapse-tools/pull/19/commits/b1c493ccdfcd4d8e9f5d955237887eb851cdce19). However, now I see a different error:
```
Successfully installed synapse-tools-0.11.5
make[1]: Leaving directory `/work/src'
   dh_installinit
   dh_installmenu
   dh_installmime
   dh_installmodules
   dh_installlogcheck
   dh_installlogrotate
   dh_installpam
   dh_installppp
   dh_installudev
   dh_installwm
   dh_installxfonts
   dh_installgsettings
   dh_bugfiles
   dh_ucf
   dh_lintian
   dh_gconf
   dh_icons
   dh_perl
   dh_usrlocal
   dh_link
   dh_compress
   dh_fixperms
   dh_strip
BFD: debian/synapse-tools/opt/venvs/synapse-tools/lib/python2.7/site-packages/st8bhKyl: Not enough room for program he
aders, try linking with -N
strip:debian/synapse-tools/opt/venvs/synapse-tools/lib/python2.7/site-packages/st8bhKyl[.note.gnu.build-id]: Bad value
BFD: debian/synapse-tools/opt/venvs/synapse-tools/lib/python2.7/site-packages/st8bhKyl: Not enough room for program he
aders, try linking with -N
strip:debian/synapse-tools/opt/venvs/synapse-tools/lib/python2.7/site-packages/st8bhKyl: Bad value
dh_strip: strip --remove-section=.comment --remove-section=.note --strip-unneeded debian/synapse-tools/opt/venvs/synap
se-tools/lib/python2.7/site-packages/_cffi_backend.so returned exit code 1
make: *** [binary] Error 9
dpkg-buildpackage: error: debian/rules binary gave error exit status 2
ERROR: InvocationError: '/nail/home/avadhutp/pg/synapse-tools/.tox/package_lucid/bin/fig run lucid'
package_lucid runtests: commands[2] | fig stop
package_lucid runtests: commands[3] | fig rm --force
No stopped containers
______________________________________________________ summary _______________________________________________________
ERROR:   package_lucid: commands failed
make: *** [package_lucid] Error 1
```

dockerfiles/lucid/Dockerfile
```diff
@@ -17,4 +18,8 @@ RUN apt-get -y install \
     build-essential \
     dh-virtualenv
 
+RUN pip install -U pip
```
solarkennedy: can you add a comment as to why this is needed?

dockerfiles/lucid/Dockerfile
```diff
@@ -17,4 +18,8 @@ RUN apt-get -y install \
...
+RUN /usr/bin/update-alternatives --install /usr/bin/pip pip /usr/local/bin/pip 3
+RUN pip install -U tox
```
solarkennedy: Is the system package of tox not enough?

If not then can you remove line 8?

solarkennedy:
@avadhutp I've never seen those errors before, my best guess is to pin dh-virtualenv back to a previously known working version to see if they go away.

avadhutp:
@solarkennedy Weird, the build passes on travis 😖 . Anyway, I will just proceed with the testing using example_happyhour for now.

solarkennedy:
It isn't weird. Travis is public and only tests trusty. Lucid is internal and uses our own docker image, pypi, etc. They are two very different build environments.

avadhutp:
@solarkennedy Tried with all versions of dh-virtualenv in internal apt (0.11-1, 0.8-1, 0.6-yelp, 0.5-1). But the same error persists. Just refuses to build on lucid.

solarkennedy:
Alright then maybe we need some help from @asottile.

```BFD: debian/synapse-tools/opt/venvs/synapse-tools/lib/python2.7/site-packages/st8bhKyl: Not enough room for program he
aders, try linking with -N
strip:debian/synapse-tools/opt/venvs/synapse-tools/lib/python2.7/site-packages/st8bhKyl[.note.gnu.build-id]: Bad value
BFD: debian/synapse-tools/opt/venvs/synapse-tools/lib/python2.7/site-packages/st8bhKyl: Not enough room for program he
aders, try linking with -N
strip:debian/synapse-tools/opt/venvs/synapse-tools/lib/python2.7/site-packages/st8bhKyl: Bad value
dh_strip: strip --remove-section=.comment --remove-section=.note --strip-unneeded debian/synapse-tools/opt/venvs/synap
se-tools/lib/python2.7/site-packages/_cffi_backend.so returned exit code 1
make: *** [binary] Error 9
dpkg-buildpackage: error: debian/rules binary gave error exit status 2```

@avadhutp for full context can you put the full lucid build output in a pb?

@asottile Is updating pip an appropriate thing to do inside the dockerfile itself?
And we still have this:
https://github.com/Yelp/synapse-tools/blob/82a8884ca191d4e08268a8127bc30f5f549838ef/src/debian/rules#L12
So I don't *think* we are using wheels on the lucid build, but those weird BFD errors make me think that it is trying to use a wheel from a future distro that it can't understand the binary?

But I don't trust my guessing. Can you advise us?

**Comment by avadhutp** [comment]
@solarkennedy @asottile Thanks to @keymone, I think I found the problem. The issue is probably this https://github.com/spotify/dh-virtualenv/issues/150. Will update more once I know for sure and fix it.

**Comment by asottile** [comment]
This is due to manylinux wheels.  You can disable these by using our
internal pypi or by specifying --no-binary :all: (or the deprecated
--no-use-wheel)

**Comment by asottile** [comment]
Another option if you're using dh-virtualenv 1.x is to use `--preinstall=no-manylinux1` which'll install https://github.com/asottile/no-manylinux1

File: dockerfiles/lucid/Dockerfile (hunk 2 above)
Comment (asottile) [inline dockerfiles/lucid/Dockerfile]: This is probably a bad idea, why do you need this?

File: dockerfiles/lucid/Dockerfile (hunk 3 above)
Comment (asottile) [inline dockerfiles/lucid/Dockerfile]: This seems unnecessary, you're not actually changing anything to do with tox

File: dockerfiles/lucid/Dockerfile (hunk 2 above)
Comment (avadhutp) [inline dockerfiles/lucid/Dockerfile]: @asottile Without this, I am seeing this error:
```
Downloading/unpacking py>=1.4.29 (from pytest)
  Running setup.py (path:/work/src/.tox/lucid/build/py/setup.py) egg_info for package py
    
    warning: no files found matching 'README.txt'
Downloading/unpacking setuptools from https://pypi.python.org/packages/64/88/d434873ba1ce02c0ed669f574afeabaeaaeec207
929a41b5c1ed947270fc/setuptools-34.1.0.zip#md5=a2f819c503b49781b1a15cfb4d49c974 (from pytest)
  Running setup.py (path:/work/src/.tox/lucid/build/setuptools/setup.py) egg_info for package setuptools
    Traceback (most recent call last):
      File "<string>", line 3, in <module>
      File "setuptools/__init__.py", line 10, in <module>
        from six.moves import filter, map
    ImportError: No module named six.moves
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):

  File "<string>", line 3, in <module>

  File "setuptools/__init__.py", line 10, in <module>

    from six.moves import filter, map

ImportError: No module named six.moves

```

File: dockerfiles/lucid/Dockerfile (hunk 3 above)
Comment (avadhutp) [inline dockerfiles/lucid/Dockerfile]: @asottile Without this, I am seeing this error
```
Downloading/unpacking py>=1.4.29 (from pytest)
  Running setup.py (path:/work/src/.tox/lucid/build/py/setup.py) egg_info for package py
    
    warning: no files found matching 'README.txt'
Downloading/unpacking setuptools from https://pypi.python.org/packages/64/88/d434873ba1ce02c0ed669f574afeabaeaaeec207
929a41b5c1ed947270fc/setuptools-34.1.0.zip#md5=a2f819c503b49781b1a15cfb4d49c974 (from pytest)
  Running setup.py (path:/work/src/.tox/lucid/build/setuptools/setup.py) egg_info for package setuptools
    Traceback (most recent call last):
      File "<string>", line 3, in <module>
      File "setuptools/__init__.py", line 10, in <module>
        from six.moves import filter, map
    ImportError: No module named six.moves
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):

  File "<string>", line 3, in <module>

  File "setuptools/__init__.py", line 10, in <module>

    from six.moves import filter, map

ImportError: No module named six.moves
```

File: dockerfiles/lucid/Dockerfile (hunk 2 above)
Comment (asottile) [inline dockerfiles/lucid/Dockerfile]: You're not pinning everything!  You should be pinning setuptools.  See https://github.com/Yelp/requirements-tools

Comment (asottile) [inline dockerfiles/lucid/Dockerfile]: @solarkennedy the fix in paasta is not ideal either and we probably shouldn't be copy pasting it to more projects

File: src/synapse_tools/configure_synapse.py [hunk 4]
```diff
@@ -252,12 +269,15 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
 
             synapse_config['services'][backend_identifier] = config
 
+        proxied_through = service_info.get('proxied_through')
```
solarkennedy: Won't this result in a key error all the time? Maybe make the default None ?

asottile: in python, `.get()` defaults to `None`

src/synapse_tools/configure_synapse.py
```diff
@@ -169,8 +169,33 @@ def get_backend_name(service_name, discover_type, advertise_type):
...
     frontend_acl_configs = []
+
+    # check for proxied_through first, use_backend ordering matters
+    if proxied_through:
+        frontend_acl_configs.extend(
+            [
+                'acl is_status_request path {healthcheck_uri}'.format(
+                    healthcheck_uri=healthcheck_uri,
+                ),
+                'acl request_from_proxy hdr_beg(X-SmartStack-Proxied_through) -i {proxied_through}'.format(
```
ArnaudBrousseau: Talked with @avadhutp about this. In a followup change we'll probably want to:

* call this `X-Yelp-Source` (or `X-Smartstack-Source` if we want to be friendlier towards non-Yelp user of Yelp/synapse-tools...I don't have a strong opinion).
* configure HAProxy to automatically add `X-Yelp-Source: caller.nerve.namespace` when a call is first proxied
* configure HAProxy to automatically add `X-Yelp-Destination: destination.nerve.namespace` when a call is first proxied. This will let proxy_service know where to forward the request to (path of a request: caller->HAProxy->proxy_service->HAProxy->service)

src/tests/configure_synapse_test.py
```diff
@@ -320,6 +320,92 @@ def test_generate_configuration_empty(mock_available_location_types):
...
+                    'extra_headers': {
+                        'X-Mode': 'ro'
+                    },
+                    'extra_healthcheck_headers': {
+                        'X-Mode': 'ro'
+                    },
+                    'balance': 'roundrobin',
+                    'advertise': ['region'],
+                    'discover': 'region',
+                    'proxied_through': 'spectre.main',
```
ArnaudBrousseau: modify to proxy_service.main?

src/tests/configure_synapse_test.py
```diff
@@ -320,6 +320,92 @@ def test_generate_configuration_empty(mock_available_location_types):
...
+                    'timeout client 3000ms',
+                    'bind /var/run/synapse/sockets/test_service.sock',
+                    'capture request header X-B3-SpanId len 64',
+                    'capture request header X-B3-TraceId len 64',
+                    'capture request header X-B3-ParentSpanId len 64',
+                    'capture request header X-B3-Flags len 10',
+                    'capture request header X-B3-Sampled len 10',
+                    'option httplog',
+                    'acl is_status_request path /status',
+                    'acl request_from_proxy hdr_beg(X-SmartStack-Proxied_through) -i spectre.main',
```
ArnaudBrousseau: that'll have to change to say "proxy_service" instead of "spectre" too :)

avadhutp:
@ArnaudBrousseau So, I've added two new headers to the requests now
1. `X-Smartstack-Source`: This had to be added to the backend of the service being proxied to. This is because, if we add it to the frontend of the calling service, then the acl to detect the headers evaluates to true and does not use the proxy service's backend.
2. `X-Smartstack-Destination`: This was pretty straightforward—added to the frontend of the calling service.

I added some temporary logging into Spectre and these are the headers I see on every request (except `status`):
```
host:dev8-devc:20074
x-smartstack-destination:example_happyhour.main
accept:*/*
x-smartstack-source:spectre.main
connection:close
user-agent:curl/7.19.7 (x86_64-pc-linux-gnu) libcurl/7.19.7 OpenSSL/0.9.8k zlib/1.2.3.3 libidn/1.23
```

I also modified the unit tests to assert this behavior and incorporated your other suggestions.

drolando:
I can't update the CR directly, so I'll just paste the diff here: http://pastebin.com/MbtFAaPN

setuptools>=34 can't be installed from source anymore (https://setuptools.readthedocs.io/en/latest/history.html#v34-0-0). You either need to use the wheel or the system package. We explicitly forbid wheels and we run `pip install --upgrade` with an unpinned setuptools, which makes pip download the latest version and try to build it from source.

The easiest solution is just to pin setuptools to 33.1.1 until we can get rid of lucid. With this patch both lucid's and trusty's builds and itests pass. I haven't tried with xenial but I don't see why it should fail.

asottile:
Another solution is to allow wheels but forbid the manylinux wheels which cause dh-virtualenv trouble.  You can do this with --preinstall no-manylinux1

This will install https://github.com/asottile/no-manylinux1

avadhutp:
All done. Builds fixed too.

avadhutp:
@ArnaudBrousseau @solarkennedy I don't have push access to this repo. Can one of you merge this upstream as `v0.12.0`?

ArnaudBrousseau:
@avadhutp I'd like to have @solarkennedy or somebody else from ops sign off on this change first.

PR #20, opened 2017-02-24
Adding errorfiles options

cblkwell:
This change adds the ability to specify different default errorfiles in the haproxy-synapse config; it should be entirely optional.

### DIFF
--- a/src/synapse_tools/configure_synapse.py
+++ b/src/synapse_tools/configure_synapse.py
@@ -159,6 +159,31 @@ def generate_base_config(synapse_tools_config):
             }
         }
     }
+
+    # This allows us to add optional non-default error file directives; they
+    # should be a nested JSON object within the synapse-tools config of this
+    # sort:
+    # {
+    #   'errorfiles': {
+    #     '404': '/etc/haproxy-synapse/errors/404.http',
+    #     '503': '/etc/haproxy-synapse/errors/503.http'
+    #   }
+    # }
+    #
+    # This will add the following lines to the 'defaults' section of the
+    # haproxy config:
+    # errorfile 404 /etc/haproxy-synapse/errors/404.http
+    # errorfile 503 /etc/haproxy-synapse/errors/503.http
+    if synapse_tools_config['errorfiles']:
+        error_list = []
+        for error in synapse_tools_config['errorfiles']:
+            new_error = 'errorfile ' + error + ' ' + synapse_tools_config['errorfiles'][error]
+            error_list.append(new_error)
+
+
+        base_config['defaults'].extend(new_error)
+
+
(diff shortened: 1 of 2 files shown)


src/synapse_tools/configure_synapse.py
```diff
@@ -159,6 +159,29 @@ def generate_base_config(synapse_tools_config):
...
+    #     '404': '/etc/haproxy-synapse/errors/404.http',
+    #     '503': '/etc/haproxy-synapse/errors/503.http'
+    #   }
+    # }
+    #
+    # This will add the following lines to the 'defaults' section of the
+    # haproxy config:
+    # errorfile 404 /etc/haproxy-synapse/errors/404.http
+    # errorfile 503 /etc/haproxy-synapse/errors/503.http
+    if synapse_tools_config.get('errorfiles', False):
```
Rob-Johnson: how about:

```
errors = [
   "errorfile %s %s".format(error, errorfile)
    for error, errorfile in synapse_tools_config.get('errorfiles', []).iteritems()
]
base_config['haproxy']['defaults'].extend(error_list)
```

cblkwell: If errorfiles is not defined, will that still be okay? I guess it should, but my python is still...iffy. If that's good then I have no problem with the more elegant solution.

Rob-Johnson: yeah, the second argument to ``synapse_tools_config.get`` is a default

```
robj@robj-MacBookPro:~/workspace/chronos master % python
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> x = {}
>>> x.get('foo', [])
[]
```

Rob-Johnson: oh, wait, you'll want the default argument to be a dictionary rather than a list:

```
errors = [
   "errorfile %s %s".format(error, errorfile)
    for error, errorfile in synapse_tools_config.get('errorfiles', {}).iteritems()
]
base_config['haproxy']['defaults'].extend(error_list)
```

cblkwell: So, that code doesn't appear to actually work -- when I run the tests, I'm getting "errorfile %s %s" getting added to the list, not the actual output.  Not sure what I'm missing here, exactly, but I suspect you can probably spot the issue faster than I will figure it out, so I thought I'd mention it. :)

Rob-Johnson: ugh, sorry. it should be ``"errorfile {} {}".format(error, errorfile)``

PR #25, opened 2017-04-11
Fixes for nginx listeners

jolynch:
This addresses the issues with nginx listeners. Specifically:

- Got rid of user directive since we're not launching as root
- Made reloads use the "graceful reload" strategy instead of just HUP. This way new binaries will automatically be picked up
- Cleaned up timeouts to just be the same high value for all listeners (60 minutes since that's or max SS timeout). This means that the only thing we will restart nginx for is new services (this happens basically never)
- Rate limit nginx restarts to once every 10 minutes (since new services rarely happen)
- Try to make the headers less wonky (unfortunately no way to kill Date or Server, but we can at least pass backing Servers forward).

### DIFF
--- a/src/synapse_tools/configure_synapse.py
+++ b/src/synapse_tools/configure_synapse.py
@@ -57,6 +57,8 @@ def set_defaults(config):
         ('nginx_prefix', '/var/run/synapse/nginx_temp'),
         ('nginx_config_path', '/var/run/synapse/nginx.cfg'),
         ('nginx_pid_file_path', '/var/run/synapse/nginx.pid'),
+        # http://nginx.org/en/docs/control.html#upgrade
+        # This is apparently how you gracefully reload the binary ...
         ('nginx_reload_cmd_fmt',
             'kill -USR2 $(cat ${nginx_pid_file_path}) && sleep 2 && '
             'kill -WINCH $(cat ${nginx_pid_file_path}.oldbin) && '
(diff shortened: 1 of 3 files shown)


jolynch:
@jnb this is for SMTSTK-182

src/synapse_tools/configure_synapse.py
```diff
@@ -611,25 +617,19 @@ def _generate_haproxy_for_watcher(service_name, service_info, synapse_tools_conf
 def _generate_nginx_for_watcher(service_name, service_info, synapse_tools_config):
     socket_path = _get_socket_path(synapse_tools_config, service_name)
 
-    # For the nginx listener, we just want the highest possible timeout
-    timeout = max(
-        service_info.get('timeout_client_ms'),
-        service_info.get('timeout_server_ms'),
-    )
+    # For the nginx listener, we just want the highest possible timeout.
+    # To limit restarts we set this to the max reap age (so HAProxy will
```
jnb: Should this be the max reap age plus epsilon?

src/synapse_tools/configure_synapse.py
```diff
@@ -89,14 +91,15 @@ def _generate_nginx_top_level(synapse_tools_config):
...
-                'user {0}'.format(synapse_tools_config['nginx_user_group']),
                 'error_log /dev/null crit',
             ],
             'http': [
                 'access_log off',
                 'sendfile on',
                 'tcp_nopush on',
                 'tcp_nodelay on',
+                'server_tokens off',
+                'proxy_pass_header Server',
```
jnb: Are there any other headers that we should be treating specially?

src/synapse_tools/configure_synapse.py
```diff
@@ -57,13 +58,14 @@ def set_defaults(config):
         ('nginx_config_path', '/var/run/synapse/nginx.cfg'),
         ('nginx_pid_file_path', '/var/run/synapse/nginx.pid'),
         ('nginx_reload_cmd_fmt',
-            '{nginx_path} -s reload -c {nginx_config_path} -p {nginx_prefix}'),
+            'kill -USR2 $(cat ${nginx_pid_file_path}) && sleep 2 && '
+            'kill -WINCH $(cat ${nginx_pid_file_path}.oldbin) && '
```
jnb: What's up with .oldbin?

jolynch: That's actually how nginx does graceful reloads :-/

http://nginx.org/en/docs/control.html#upgrade

I hate this.

src/synapse_tools/configure_synapse.py
jolynch: I don't think so. I think by default it passes any upstream headers, it's just the Server header that nginx decides to add no matter what.

src/synapse_tools/configure_synapse.py
jolynch: sure, /me shrug

PR #27, opened 2017-04-17
Tries to make NGINX a more proper interactive proxy

jolynch:
A lot of this is cribbed from [k8ns](https://github.com/kubernetes/ingress/blob/master/controllers/nginx/rootfs/etc/nginx/template/nginx.tmpl) and nginx docs on [reverse proxying](https://www.nginx.com/resources/admin-guide/reverse-proxy/) and [websockets](http://nginx.org/en/docs/http/websocket.html). Doing some manual testing indicates that the proper headers are passed now.

I increased the buffers because HAProxy has [16k buffers](https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#3.2-tune.bufsize) instead of 8k by default.

proxy_max_temp_file_size is set to 0 to prevent any buffering of responses to files (wth why would you ever do that :-( )

More [sources](https://serverfault.com/questions/768693/nginx-how-to-completely-disable-request-body-buffering).

@jnb @EvanKrall heads up

### DIFF
--- a/dockerfiles/itest/itest/synapse-tools-both.conf.json
+++ b/dockerfiles/itest/itest/synapse-tools-both.conf.json
@@ -4,5 +4,7 @@
   "stats_port": 32123,
   "haproxy.defaults.inter": "10s",
   "listen_with_nginx": true,
-  "listen_with_haproxy": true
+  "listen_with_haproxy": true,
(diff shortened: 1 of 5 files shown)


jolynch:
@vulpine if you see anything i'm missing please let me know

PR #28, opened 2017-05-18
Create /var/run/synapse/sockets in configure_synpase

jglukasik:


src/synapse_tools/configure_synapse.py
```diff
@@ -457,6 +463,8 @@ def main():
         my_config, get_zookeeper_topology(my_config['zookeeper_topology_path']), get_all_namespaces()
     )
 
+    ensure_directory_exists('/var/run/synapse/sockets')
```
jolynch: For /var/run/synapse/services we take care of this in the Synapse init script (see puppet modules/synapse_simple/templates/synapse-run.erb). I sort of like that there, Synapse can't run without its init script.

src/synapse_tools/configure_synapse.py
```diff
@@ -449,6 +449,12 @@ def get_my_grouping(grouping_type):
     with open('/nail/etc/{0}'.format(grouping_type)) as fd:
         return fd.read().strip()
 
+def ensure_directory_exists(path):
+    try:
+        os.makedirs(path)
```
jolynch: if we do keep this here we probably need to chown to nobody

jglukasik:
Ahh, yeah that init script option seems better here. Closing this.

PR #31, opened 2017-07-14
Make HAProxy listen on PROXY protocol sockets

jvperrin:
Nginx is still sending to the same sockets as before though, so nothing has changed on that front yet, but I think that switch would be tied to the `proxy_proto` config variable.


src/synapse_tools/configure_synapse.py
```diff
@@ -55,6 +57,7 @@ def set_defaults(config):
             '/nail/etc/zookeeper_discovery/infrastructure/local.yaml'),
         ('hacheck_port', 6666),
         ('stats_port', 3212),
+        ('proxy_proto', False),
```
jolynch: nit: I'd group this with the other haproxy options and have ``haproxy_proxy_bind`` or something as the name

PR #32, opened 2017-07-18
Fix build for lucid and xenial

jvperrin:
This also adds xenial to the tests run by travis, but should run in parallel to keep things speedy.


solarkennedy:
Thanks!

PR #34, opened 2017-07-19
Configure synapse plugins

zareenc:
Support for Python plugins for configuring synapse, with initial support and testing for Lua logging plugin

cc: @jolynch


dockerfiles/itest/itest/Dockerfile.trusty
```diff
@@ -20,13 +20,28 @@ RUN apt-get update && apt-get -y install \
...
+ADD http://www.lua.org/ftp/lua-5.3.1.tar.gz /lua-5.3.1.tar.gz
+RUN tar -axvf /lua-5.3.1.tar.gz
+WORKDIR /lua-5.3.1
+RUN pwd
+RUN make generic
+RUN make install INSTALL_TOP=/usr/bin/lua
+
 # Ubuntu trusty nginx and haproxy are ancient, grab newer ones
+WORKDIR /
 ADD https://github.com/haproxy/haproxy/archive/v1.6.0.tar.gz /haproxy.tar.gz
```
jvperrin: Does this need to be version `1.7.8` to match the newest version of `haproxy-synapse` and have better Lua support?

dockerfiles/itest/itest/Dockerfile.trusty
```diff
@@ -20,13 +20,28 @@ RUN apt-get update && apt-get -y install \
     build-essential \
     zlib1g-dev
 
+# Lua 5.3 for scripting with HAProxy
```
jvperrin: I think you'd want to do a similar thing for the xenial itests, but there it'll be easier since the Lua 5.3 packages already exist in the public package repos.

src/synapse_tools/configure_synapse.py
```diff
@@ -475,6 +477,31 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
                     )
                 )
 
+            # Add HAProxy options if logging is enabled
+            plugins = service_info.get('plugins', {})
+            if plugins.get('logging') is not None:
```
jvperrin: Could this be `if plugins.get('logging'):` instead? (if it's not an empty list or anything "Falsy")

It would be a little cleaner if so.

src/synapse_tools/configure_synapse.py
```diff
@@ -475,6 +477,31 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
...
+                )
+                synapse_config['services'][service_name]['haproxy']['backend'].extend(
+                    ProvidenceLogging().backend_opts(service_name, service_info)
+                )
+                synapse_config['haproxy']['global'].extend(
+                    ProvidenceLogging().global_opts(service_name, service_info)
+                )
+
+            # Add HAProxy options if path based routing is enabled
+            if plugins.get('path_based_routing') is not None:
```
jvperrin: Same as above

src/synapse_tools/configure_synapse.py
```diff
@@ -475,6 +477,31 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
...
+                synapse_config['services'][service_name]['haproxy']['frontend'].extend(
+                    PathBasedRouting().frontend_opts(service_name, service_info)
+                )
+                synapse_config['services'][service_name]['haproxy']['backend'].extend(
+                    PathBasedRouting().backend_opts(service_name, service_info)
+                )
+                synapse_config['haproxy']['global'].extend(
+                    PathBasedRouting().global_opts(service_name, service_info)
+                )
+
```
jvperrin: These two sections have a fair amount of duplication, it seems to me that they could potentially be combined into a more generic plugin selector

dockerfiles/itest/itest/Dockerfile.trusty
jvperrin: (probably why tests are currently failing)

zareenc: Yep I think you're right, I'll add for xenial

jvperrin: Alright, cool, I think all you need is the `liblua5.3-dev` package and that should be enough to compile haproxy with from my testing with that.

jvperrin: I was just looking at this again, and although the tests don't show it here, lucid will actually need something similar to trusty and xenial. However, I think you can do the exact same as xenial (minus `libssl-dev`), since the lucid Dockerfile is using the `lucid_yelp` image. It isn't built by Travis because it is ancient and doesn't have the proper packages, but it's used to upload the synapse-tools package internally, so it does need to be able to build there at least otherwise the package won't build and then can't be uploaded until it builds for lucid too.

zareenc: Discussed with @jvperrin, and we realized we don't actually need to add Lua to lucid's Dockerfile because it already includes `apt-get install haproxy-synapse`, and haproxy-synapse is now built with Lua 5.3. I did have to add the lua scripts to lucid's Dockerfile, though, for the plugin itests.

dockerfiles/itest/itest/itest.py
```diff
@@ -160,6 +172,10 @@ def _sort_lists_in_dict(d):
     return d
 
 
+def test_haproxy_config_valid(setup):
```
jolynch: ++

dockerfiles/itest/itest/lua_scripts/log_requests.lua
```diff
@@ -0,0 +1,38 @@
...
+  if hdr["from"] ~= nil then
+    from = hdr["from"][0]
+  end
+
+  local log_text = 'Source: ' .. ip .. ' From: ' .. from .. '\n'
+  txn.Info(txn, log_text)
+
+  local fname = LOGFILE_PATH .. 'demo_log'
+  local log_file = io.open(fname, 'a')
+  log_file:write(log_text)
```
jolynch: Since this is going to a file could it block potentially?

Does HAProxy have a log api or other udp system which would block less likely? Or is this running in the "background" task?

src/synapse_tools/config_plugins/HAProxyConfigPlugin.py
```diff
@@ -0,0 +1,19 @@
+import abc
+
+LUA_SCRIPTS_PATH = '/nail/etc/lua_scripts/'
+
+
+class HAProxyConfigPlugin(object):
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def global_opts(self, service_name, service_info):
```
jolynch: nitpick: in synapse we used to abbreviate options to opts everywhere and then we ended up with args, opts, and options. Let's just go with options or lines always?

e.g. ``global_options`` or ``globale_lines``? I don't feel strongly about this just a thought.

src/synapse_tools/config_plugins/HAProxyConfigPlugin.py
```diff
@@ -0,0 +1,19 @@
+import abc
```
jolynch: generally don't name files with uppercase.

I'd recommend something like ``base.py`` or ``interface.py``

src/synapse_tools/config_plugins/HAProxyConfigPlugin.py
```diff
@@ -0,0 +1,19 @@
...
+
+LUA_SCRIPTS_PATH = '/nail/etc/lua_scripts/'
```
jolynch: This should probably be a configuration option to configure_synapse or if the plugin requires lua then an argument to the constructor. Does it need to be in the base?

src/synapse_tools/config_plugins/HAProxyConfigPlugin.py
```diff
@@ -0,0 +1,19 @@
...
+
+class HAProxyConfigPlugin(object):
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def global_opts(self, service_name, service_info):
+        return
+
+    @abc.abstractmethod
+    def frontend_opts(self, service_name, service_info):
```
jolynch: Also can you leave docstrings that explain what the arguments are and what these should return (mostly type information is useful)

src/synapse_tools/config_plugins/PathBasedRouting.py
```diff
@@ -0,0 +1,16 @@
+from HAProxyConfigPlugin import HAProxyConfigPlugin
+from HAProxyConfigPlugin import LUA_SCRIPTS_PATH
+
+
+class PathBasedRouting(HAProxyConfigPlugin):
+    def global_opts(self, service_name, service_info):
+        return ['lua-load %spath_based_routing.lua' % LUA_SCRIPTS_PATH]
+
+    def frontend_opts(self, service_name, service_info):
```
jolynch: So I don't think we need to fix this now, but how are we going to enforce ordering in the use_backend lines? afaik the chains can only be linear, so the order that the plugins write these out will matter. I think we can fix this in the next iteration but it might be a good idea to think about that edge case a bit.

src/synapse_tools/configure_synapse.py
```diff
@@ -17,6 +17,14 @@
 from paasta_tools.marathon_tools import get_all_namespaces
 from synapse_tools.haproxy_synapse_reaper import DEFAULT_REAP_AGE_S
 from yaml import CLoader
+from synapse_tools.config_plugins.ProvidenceLogging import ProvidenceLogging
+from synapse_tools.config_plugins.PathBasedRouting import PathBasedRouting
+
+
+PLUGIN_MAP = {
```
jolynch: I'd put this in the plugins directory somewhere, probably in ``base.py``.

src/synapse_tools/configure_synapse.py
```diff
@@ -475,6 +483,21 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
                     )
                 )
 
+            # Add HAProxy options for plugins
+            plugins = service_info.get('plugins', {})
+            for plugin_name in sorted(PLUGIN_MAP.keys()):
```
jolynch: What is instead of sorting we used an ordered dict or something to define precedence of plugins? Since the ACL ordering might be important? What do you think?

src/synapse_tools/configure_synapse.py
```diff
@@ -475,6 +483,21 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
...
+                if plugins.get(plugin_name):
+                    plugin_class = PLUGIN_MAP[plugin_name]
```
jolynch: nit: you're not really retrieving the class here, but an instance of the class, maybe make it a bit clearer with ``plugin_instance`` or something?

src/synapse_tools/config_plugins/PathBasedRouting.py
```diff
@@ -0,0 +1,16 @@
+from HAProxyConfigPlugin import HAProxyConfigPlugin
+from HAProxyConfigPlugin import LUA_SCRIPTS_PATH
+
+
+class PathBasedRouting(HAProxyConfigPlugin):
+    def global_opts(self, service_name, service_info):
+        return ['lua-load %spath_based_routing.lua' % LUA_SCRIPTS_PATH]
```
jolynch: This lua script isn't in the diff is it? Do we want to leave this file out (bring it in the next iteration) or include the lua?

src/synapse_tools/configure_synapse.py
```diff
@@ -475,6 +483,21 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
                     )
                 )
 
+            # Add HAProxy options for plugins
+            plugins = service_info.get('plugins', {})
```
jolynch: Also keep in mind that we have to thread this new option through service_configuration_lib pre-receive hooks and paasta_tools. kwa should have context for which pieces need to change here.

src/synapse_tools/config_plugins/PathBasedRouting.py
zareenc: Nope, I decided to exclude the actual Lua script since we are still missing some of the pieces (e.g. mapping from services to IP, where exactly to send logs). I wanted to hold off until we get those parts figured out. However I quickly wrote the itest logging script just for testing purposes (since it was just for testing, I didn't optimize for running it on a background thread).

src/synapse_tools/config_plugins/HAProxyConfigPlugin.py
zareenc: Adding it as an entry in `defaults` and making the plugin classes inherit from a LuaPlugin superclass whose constructor takes the path as a parameter

src/synapse_tools/config_plugins/HAProxyConfigPlugin.py
zareenc: Refactoring plugin files to `base.py`, `logging.py`, and `path_based_routing.py`

src/synapse_tools/config_plugins/HAProxyConfigPlugin.py
zareenc: Sure, going with `global_options`

src/synapse_tools/config_plugins/HAProxyConfigPlugin.py
zareenc: Good idea, adding those in

src/synapse_tools/configure_synapse.py
zareenc: renamed

src/synapse_tools/configure_synapse.py
zareenc: Putting this in a separate file `map.py` to avoid circular imports (`base.py` would have to import `logging.py` which imports `base.py`)

src/synapse_tools/configure_synapse.py
zareenc: Good idea, using an OrderedDict for the map to establish prioritization among the plugins. Tests need to follow this ordering as well.

src/synapse_tools/configure_synapse.py
zareenc: Adding plugin key to smartstack json schema in `py-gitolite`

dockerfiles/itest/itest/lua_scripts/log_requests.lua
zareenc: Keeping this for now in the testing script, will investigate background tasks/publishing to Scribe for the actual logging script

src/synapse_tools/configure_synapse.py
solarkennedy: Currently we don't whitelist the keys here. You are free to add keys to get you going, but bonus points for adding validation too. I gave you a code hint on irc.

src/synapse_tools/config_plugins/base.py
```diff
@@ -0,0 +1,43 @@
...
+        Options for HAProxy configuration frontend section
+        :param str service_name: name of service
+        :param dict service_info: dictionary of service config info
+        :return: list of strings representing distinct
+                 lines in HAProxy config frontend
+        """
+        return
+
+    @abc.abstractmethod
+    def backend_options(self, service_name, service_info):
```
jolynch: I think these need to take the synapse_tools config in addition to the service_info.

Also is the ``service_name`` redundant with the service_info hash? Just wondering.

src/synapse_tools/config_plugins/base.py
```diff
@@ -0,0 +1,43 @@
...
+        Options for HAProxy configuration backend section
+        :param str service_name: name of service
+        :param dict service_info: dictionary of service config info
+        :return: list of strings representing distinct
+                 lines in HAProxy config backend
+        """
+        return
+
+
+class LuaPlugin(object):
```
jolynch: If the functions above take the synapse-tools config then I don't think we need this (lua_directory can be a configuration option)

src/synapse_tools/configure_synapse.py
```diff
@@ -491,6 +494,27 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
...
+            for plugin_name in PLUGIN_MAP:
+                if plugin_name not in plugins:
+                    continue
+                if issubclass(PLUGIN_MAP[plugin_name], LuaPlugin):
```
jolynch: Let's avoid this kind of subclass switching if possible. I think you can nix the LuaPlugin interface entirely and avoid this (set the lua_dir in the synapse-tools option to be this path by default and then if folks want to move the lua scripts to e.g. be puppet managed they can get it from there)

src/synapse_tools/configure_synapse.py
```diff
@@ -491,6 +494,27 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
...
+                    lua_dir_path = os.path.join(os.path.dirname(synapse_tools.__file__), 'lua_scripts')
+                    plugin_instance = PLUGIN_MAP[plugin_name](lua_dir_path)
+                else:
+                    plugin_instance = PLUGIN_MAP[plugin_name]()
```
jolynch: What if we just always construct the plugin with the contents of the plugin config dict? Feel free to TODO this if it's too annoying of a refactor. But I do think this would be a clean(er) interface.

E.g. 
```
PLUGIN_MAP[plugin_name](
  **plugins[plugin_name]
)
```

src/synapse_tools/configure_synapse.py
jvperrin: I agree, I was talking about this during the hackathon with some people interested in this project and we agreed that moving the scripts to be in the puppet repo and then having `synapse-tools` pull all available scripts from a directory would be much easier to deploy for anyone wanting to make a new plugin or more quickly iterate on an existing plugin. This would mean that HAProxy configuration snippets would have to be managed somehow though, because each script could have separate global, frontend, and backend options that would need to be dynamically loaded...

src/synapse_tools/config_plugins/base.py
```diff
@@ -0,0 +1,32 @@
+import abc
+
+
+class HAProxyConfigPlugin(object):
+    __metaclass__ = abc.ABCMeta
+
```
jolynch: Let's do an init method here to enforce the interface.

PR #41, opened 2017-08-14
Sampled logging

zareenc:
Sampling added to provenance logging plugin. In order to allow for a configurable sampling rate, the plugin format has been modified to use suboptions - all plugins must include an 'enabled' option and  can additionally include other plugin options (e.g. sample rate). 

cc: @jolynch @jvperrin @matfra


src/synapse_tools/configure_synapse.py
```diff
@@ -499,24 +499,31 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
...
-                    (synapse_config['services'][service_name]['haproxy']['backend'],
-                     plugin_instance.backend_options()),
-                    (synapse_config['haproxy']['global'],
-                     plugin_instance.global_options())
-                ]
-                for (config, opts) in config_to_opts:
-                    config.extend([x for x in opts if x not in config])
+                # Check if plugin is enabled for this service or if global flag is set
+                svc_enabled = plugin_name in plugins and plugins[plugin_name]['enabled']
+                global_enabled = synapse_tools_config.get(plugin_name) and \
```
jolynch: let's prefer

```
synapse_tools_config.get(plugin_name, False)
```

src/synapse_tools/configure_synapse.py
```diff
@@ -499,24 +499,31 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
...
+                    synapse_tools_config.get(plugin_name)['enabled']
+
+                if svc_enabled or global_enabled:
+                    plugin_opts = plugins[plugin_name] if svc_enabled else \
```
jolynch: prefer implied continuations

```
a = (
    x if b else z
)
```

or

```
a = (
    x if b else
    z
)
```

src/synapse_tools/lua_scripts/log_requests.lua
```diff
@@ -1,38 +1,62 @@
...
 
-  local map_file = txn.f:env('map_file')
-  txn.Info(txn, 'Mapfile: ' .. map_file)
-  map = Map.new(map_file, Map.str)
+  -- Set sample rate
+  local sample_rate = txn.f:env('sample_rate')
+  if sample_rate ~= nil and not sample then
+    sample = true
+    if tonumber(sample_rate) > 0 then
+      max_count = math.floor(tonumber(sample_rate))
```
jolynch: What's the range of sample_rate? If it's 0 -> 1 does this work?

src/synapse_tools/config_plugins/logging.py
```diff
@@ -9,14 +9,18 @@ def global_options(self):
...
         map_dir = self.synapse_tools_config['map_dir']
         map_file = os.path.join(map_dir, 'ip_to_service.map')
-        return [
+        opts = [
             'lua-load %s' % lua_file,
             'setenv map_file %s' % map_file
         ]
+        if 'sample_rate' in self.plugin_opts:
+            sample_rate = str(self.plugin_opts['sample_rate'])
+            opts.append('setenv sample_rate %s' % sample_rate)
```
jolynch: I'd just use new style substitutions here and not worry about types.

```
opts.append('setenv sample_rate {0}'.format(sample_rate))
```

src/synapse_tools/lua_scripts/log_requests.lua
```diff
@@ -1,38 +1,62 @@
...
-     txn.Info(txn, 'Could not find source IP address')
      ip = 'nil'
   end
-  txn.Info(txn, 'Source ip: ' .. ip)
-
   src_svc = map:lookup(ip)
   if src_svc == nil then
-     txn.Info(txn, 'Could not find source service')
-     src_svc = 'nil'
+     src_svc = ip
```
jolynch: I'm a bit worried about this separation of functions and coupled state through log_req / src_src

Can we combine these into one function which decides to sample or not and just logs the provenance (no Source Service line)?

src/synapse_tools/config_plugins/logging.py
zareenc: Sure thing

src/synapse_tools/lua_scripts/log_requests.lua
zareenc: Good catch, looks like I forgot to add in a change from the docker container I was testing in. It should be 1/sample_rate, which is why there's a check that the sample_rate isn't 0. Will fix this.

src/synapse_tools/configure_synapse.py
zareenc: Sure

src/synapse_tools/configure_synapse.py
zareenc: Fixed

src/synapse_tools/lua_scripts/log_requests.lua
zareenc: Agree, this is a bit hacky. I'll test if the backend has access to the same source information as the frontend, then there can just be a single function call in the backend instead of two.

zareenc: Tested it out, looks like the backend does have access to request information as well. Will move source & destination logging into one function to make this cleaner.

src/synapse_tools/config_plugins/logging.py
```diff
@@ -20,22 +26,22 @@ def global_options(self):
         lua_file = os.path.join(lua_dir, 'log_requests.lua')
         map_dir = self.synapse_tools_config['map_dir']
         map_file = os.path.join(map_dir, 'ip_to_service.map')
-        return [
+        opts = [
             'lua-load %s' % lua_file,
             'setenv map_file %s' % map_file
         ]
+        if 'sample_rate' in self.plugin_opts:
+            sample_rate = str(self.plugin_opts['sample_rate'])
```
jolynch: Is sample_rate guaranteed to be defined? Perhaps we should default to something like 0 or "off" ?

src/synapse_tools/lua_scripts/log_requests.lua
```diff
@@ -1,50 +1,68 @@
...
 
-  local map_file = txn.f:env('map_file')
-  txn.Info(txn, 'Mapfile: ' .. map_file)
-  map = Map.new(map_file, Map.str)
+  -- Set sample rate
+  local sample_rate = txn.f:env('sample_rate')
+  if sample_rate ~= nil and not sample then
+    sample = true
+    if tonumber(sample_rate) > 0 then
+      max_count = math.floor(1./tonumber(sample_rate))
```
jolynch: this worries me, can we just draw a pseudo random number and compare if it is < the sample_rate?

typically prng are reasonably fast (they're just a multiply and an add typically)

src/synapse_tools/lua_scripts/log_requests.lua
```diff
@@ -1,50 +1,68 @@
...
-  txn.Info(txn, 'Source ip: ' .. ip)
-
   src_svc = map:lookup(ip)
   if src_svc == nil then
-     txn.Info(txn, 'Could not find source service')
-     src_svc = 'nil'
+     src_svc = ip
   end
-
   txn.Info(txn, 'Source service: ' .. src_svc)
```
jolynch: let's not log this, otherwise tmp_synapse_haproxy will get quite full.

PR #42, opened 2017-08-15
Refactor proxied_through to be a plugin

jolynch:
@zareenc heads up :-)


src/tests/configure_synapse_test.py
```diff
@@ -403,10 +404,10 @@ def test_generate_configuration_with_proxied_through(mock_get_current_location,
                 ],
                 'backend': [
                     'balance roundrobin',
-                    'acl is_status_request path /status',
-                    'reqadd X-Smartstack-Source:\\ proxy_service if !is_status_request',
                     'option httpchk GET /http/proxy_service/0/status',
                     'http-check send-state',
+                    'acl is_status_request path /status',
```
zareenc: Just want to make sure - the change in this ordering won't affect HAProxy behavior right?

src/synapse_tools/config_plugins/proxied_through.py
```diff
@@ -0,0 +1,48 @@
...
+            'acl is_status_request path {healthcheck_uri}'.format(
+                healthcheck_uri=healthcheck_uri,
+            ),
+            'acl request_from_proxy hdr_beg(X-Smartstack-Source) -i {proxied_through}'.format(
+                proxied_through=proxied_through,
+            ),
+            'acl proxied_through_backend_has_connslots connslots({proxied_through}) gt 0'.format(
+                proxied_through=proxied_through,
+            ),
+            'reqadd X-Smartstack-Destination:\ {service_name} if !is_status_request !request_from_proxy proxied_through_backend_has_connslots'.format(
```
jolynch: Yes, reqadd needs to be before a use_backend.

right now on prod we're seeing lots of:

```
[WARNING] 226/135406 (772352) : parsing [/var/run/synapse/haproxy.cfg:3753] : a 'reqadd' rule placed after a 'use_backend' rule will still be processed before.
```

src/synapse_tools/configure_synapse.py
```diff
@@ -498,11 +453,17 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
...
             # Add HAProxy options for plugins
             plugins = service_info.get('plugins', {})
-            for plugin_name in PLUGIN_MAP:
-                if plugin_name not in plugins and not synapse_tools_config.get(plugin_name):
+
+            for plugin_name in PLUGIN_REGISTRY:
+                # Proxied through is special because it's not
+                # controlled through the plugins config element
+                # Instead the plugin itself decides if it is enabled
+                if plugin_name == 'proxied_through':
```
jolynch: What if we just instantiated all plugins for all service watchers and they are responsible for enabling or disabling themselves?

src/synapse_tools/configure_synapse.py
```diff
@@ -518,18 +479,28 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
                 for (config, opts) in config_to_opts:
                     config.extend([x for x in opts if x not in config])
 
+            # populate the ACLs to route to the service backends, this must
+            # happen last because ordering of use_backend ACLs matters.
+            synapse_config['services'][service_name]['haproxy']['frontend'].extend(
+                generate_acls_for_service(
```
jolynch: I agree. I'll add a TODO

src/tests/configure_synapse_test.py
jolynch: It will, but it should be strictly more correct heh (before the use_backend lines were not in an order where they would work properly).

src/synapse_tools/configure_synapse.py
```diff
@@ -497,12 +452,8 @@ def generate_configuration(synapse_tools_config, zookeeper_topology, services):
...
 
             # Add HAProxy options for plugins
-            plugins = service_info.get('plugins', {})
-            for plugin_name in PLUGIN_MAP:
-                if plugin_name not in plugins and not synapse_tools_config.get(plugin_name):
-                    continue
-
-                plugin_instance = PLUGIN_MAP[plugin_name](
+            for plugin_name in PLUGIN_REGISTRY:
+                plugin_instance = PLUGIN_REGISTRY[plugin_name](
```
zareenc: ++ this is a lot cleaner

PR #43, opened 2017-08-15
Code parses in python 3

asottile:
gets the ball rolling


asottile:
(silly typo, so many parsers)

PR #45, opened 2018-01-29
PERF-2836: tarpitting requests

avadhutp:
# What?
Simply put, this is a fault-injection system at smartstack-level that will allow devs to simulate timeouts at various levels in a multi-service call. The point is, it can allow devs to trigger `RequestBudgetExceededException` on their own or dependent service/s.

# How this works?
Examples:
```
curl -I -v -H'X-Source-Id: spectre-test' -H'Host: internalapi' -H'X-Smartstack-Destination: yelp-main.internalapi' -H'X-Force-Master-Read: True' -H'X-Ctx-Tarpit: spectre.main' 'http://169.254.255.254:20678/category_yelp/?locale=en_US'
```
```
curl -I -v -H'X-Source-Id: spectre-test' -H'Host: internalapi' -H'X-Smartstack-Destination: yelp-main.internalapi' -H'X-Force-Master-Read: True' -H'X-Ctx-Tarpit: spectre.main, yelp-internalapi.main' 'http://169.254.255.254:20678/category_yelp/?locale=en_US'
```

# Testing done
Manually, on my devbox by turning off synapse cron runs and manually adding acl & tarpit rules to my `/var/run/synapse/haproxy.cfg` file.

### DIFF
--- a/src/synapse_tools/config_plugins/base.py
+++ b/src/synapse_tools/config_plugins/base.py
@@ -26,6 +26,7 @@ class HAProxyConfigPlugin(object):
         """
         return
 
+    @abc.abstractmethod
     def defaults_options(self):
         """
         Options for HAProxy configuration defaults section
--- a/src/synapse_tools/config_plugins/base.py
+++ b/src/synapse_tools/config_plugins/base.py
@@ -26,6 +26,13 @@ class HAProxyConfigPlugin(object):
         """
         return
 
+    def defaults_options(self):
+        """
+        Options for HAProxy configuration defaults section
+        :return: list of strings corresponding to distinct
+                 lines in HAProxy config defaults
+        """
+
     @abc.abstractmethod
     def frontend_options(self):
         """
--- a/src/synapse_tools/config_plugins/fault_injection.py
+++ b/src/synapse_tools/config_plugins/fault_injection.py
@@ -1,0 +1,25 @@
+from synapse_tools.config_plugins.base import HAProxyConfigPlugin
+
+
+MAX_TARPIT_TIMEOUT = '60s'
+TARPIT_HEADER = 'X-Ctx-Tarpit'
+
+
+class FaultInjection(HAProxyConfigPlugin):
+    def global_options(self):
+        return []
+
+    def defaults_options(self):
+        return ['timeout tarpit %s' % MAX_TARPIT_TIMEOUT]
+
+    def frontend_options(self):
+        return []
+
+    def backend_options(self):
+        return [
+            'acl to_be_tarpitted hdr_sub({header_name}) -i {service_name}'.format(
+                header_name=TARPIT_HEADER,
+                service_name=self.service_name,
+            ),
+            'reqtarpit . if to_be_tarpitted',
+        ]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## [v0.11.5](https://github.com/Yelp/synapse-tools/tree/v0.11.5) (2018-02-13)
+[Full Changelog](https://github.com/Yelp/synapse-tools/compare/v0.11.4...v0.11.5)
+
+**Merged pull requests:**
+
+- Enable fault-injection via the use of X-Ctx-Tarpit headers [\#45](https://github.com/Yelp/synapse-tools/pull/45) ([avadhutp](https://github.com/avadhutp))
+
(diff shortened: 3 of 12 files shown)


src/synapse_tools/config_plugins/fault_injection.py
```diff
@@ -0,0 +1,25 @@
+from synapse_tools.config_plugins.base import HAProxyConfigPlugin
+
+
+MAX_TARPIT_TIMEOUT = '60s'
```
avadhutp: This is basically how long haproxy will hold the tarpitted request for before responding with a 500.

avadhutp:
`itest_trusty` is failing for `master` too.

ArnaudBrousseau:
Nice! I'm not hugely familiar with HAProxy but at first glace it looks reasonable to me! Two small comments:
* Do we open ourselves to trouble if somebody from outsite our corporate network is able to set `X-Ctx-Tarpit`?
* Travis error'd out pretty hard on the Trusty build with:
```
pkg_resources.DistributionNotFound: The 'backports.functools_lru_cache>=1.2.1' distribution was not found and is required by arrow
```

src/synapse_tools/config_plugins/fault_injection.py
Rob-Johnson: do we not want this to be configurable?

avadhutp: @Rob-Johnson It would've been ideal but `request_budgets.yaml` has timeouts set at an endpoint level & not at a service level.

avadhutp:
@ArnaudBrousseau I am still struggling with this:

> Do we open ourselves to trouble if somebody from outsite our corporate network is able to set `X-Ctx-Tarpit`?

Any ideas? I am going to ask around in Ops too.

avadhutp:
@Rob-Johnson Any idea if it's ok if the trusty builds are failing? They're failing for `master` too.

ArnaudBrousseau:
>> Do we open ourselves to trouble if somebody from outsite our corporate network is able to set X-Ctx-Tarpit?

> Any ideas? I am going to ask around in Ops too.

A typical solution for this is to drop these headers (which we expect to be internal-only) in our extlbs. But that might affect the ease-of-use of X-Ctx-Tarpit if devs can't curl through extlbs.

We can also do IP verification in a middleware in yelp-main (similar to what we do for cprofile headers.)

Since mitigation would cost us development cycles it's worth asking how much risk we're taking by letting this be public facing (maybe that's fine?)

avadhutp:
@Rob-Johnson Please see @ArnaudBrousseau's [comment](https://github.com/Yelp/synapse-tools/pull/45#issuecomment-361558563) above. Should we be roping in anyone else from Ops to comment on this issue?

vulpine:
We could drop the X-Ctx-Tarpit header in prod only -- our load-balancer configuration scripts can tell which environment they're running in.

However, I'm curious as to whether there's a risk if users set that. If somebody adds the header and tarpits themselves, what are the potential problems? I imagine the answer is 'tying up a worker for MAX_TARPIT_TIMEOUT', which it's probably desirable to prohibit.

avadhutp:
@vulpine Wouldn't dropping the header in prod, as you suggested, mitigate this risk then?

vulpine:
@avadhutp It would indeed. Would be pretty straightforward to implement on our load-balancer configs and I don't see any issues with doing so (since we never expect or use that header at present)

EvanKrall:
Now that we've posted the names and formats of these magic headers, along with expected consequences, onto a public forum, we should expect for someone malicious to find them and try to use them. As such, we'll absolutely need to drop them at the edge.

src/synapse_tools/config_plugins/base.py
```diff
@@ -26,6 +26,13 @@ def global_options(self):
         """
         return
 
+    def defaults_options(self):
+        """
+        Options for HAProxy configuration defaults section
+        :return: list of strings corresponding to distinct
+                 lines in HAProxy config defaults
+        """
```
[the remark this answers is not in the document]
avadhutp: Missed it, added now.

avadhutp:
Fixed trusty build & also the extlb change to strip these headers to avoid misuse is also shipped out.

Can I get ship-its on this PR if there are no pending issues?

^^ @ArnaudBrousseau @vulpine @EvanKrall @Rob-Johnson

PR #49, opened 2018-02-22
Fix running simultaneous builds on a single Jenkins box

avadhutp:


avadhutp:
This fixes the problem of simultaneous builds on internal Jenkins boxes—https://jenkins-jerry.yelpcorp.com/job/synapse-tools/1551/.

PR #50, opened 2018-03-05
Add the X-Smartstack-Source header to all services

avadhutp:
This will add the Source header to all service backends and allow callers to identify themselves to the callee. We will be using this to identify calling services inside Apollo, but the applications are endless.

Post this, we can also deprecate `X-Source-Id` as having a smartstack namespace is more consistent.

We had already discussed doing this when we were enabling the `proxied_through` configuration for `Casper` but decided to control the blast radius. But now, with Apollo's need, it seems like a good time for a wider rollout.


avadhutp:
I will test this by creating a custom package and rollout gently across a couple of devboxes first and then stage, and finally prod.

tub:
Looks good to me as long as we're OK with the new header being returned on /status, as this used to be specifically excluded.

src/synapse_tools/config_plugins/proxied_through.py
```diff
@@ -34,18 +34,4 @@ def frontend_options(self):
...
     def backend_options(self):
-        # We are not a proxy for someone else
-        if not self.service_info.get('is_proxy', False):
-            return []
-
-        healthcheck_uri = self.service_info.get('healthcheck_uri', '/status')
-        service_name = self.service_name
-
-        return [
-            'acl is_status_request path {healthcheck_uri}'.format(
```
ArnaudBrousseau: I believe we should keep this acl, it's used [here](https://github.com/Yelp/synapse-tools/blob/e91bac378ef484419a5df0dea5ab32c5a769ddde/src/synapse_tools/config_plugins/proxied_through.py#L28) to conditionally forward requests to proxies.

AskattiYelp:
LG2M.

src/synapse_tools/config_plugins/proxied_through.py
avadhutp: @ArnaudBrousseau @tub 
1. This particular `acl` was only used in the **backend** section and that too only for adding the `X-Smartstack-Source` header conditionally.
1. @ArnaudBrousseau The directive to forward requests that you pointed out is in the **frontend** section and it has its own corresponding [acl](https://github.com/Yelp/synapse-tools/blob/master/src/synapse_tools/config_plugins/proxied_through.py#L19).

PR #52, opened 2018-03-21
Add X-Smartstack-Origin header for configured services

avadhutp:
# What?
It enables a plugin called `source_required`. This, when enabled, will add a `X-Smartstack-Origin` header to the request. The value of this header will be the smartstack namespace of the calling service.

# Why?
As required by DSWG, it will enable Apollo to identify clients calling it (APOLLO-457).

# How's it being tested?
I've already modified unit tests and added itest to confirm this change. But I am also cutting an rc (which is why the change in `src/setup.py`) and installing it one (or more) devbox test the impact.

### DIFF
--- a/src/synapse_tools/lua_scripts/add_source_header.lua
+++ b/src/synapse_tools/lua_scripts/add_source_header.lua
@@ -11,8 +11,7 @@ core.register_action('init_add_source', {'tcp-req', 'http-req'}, init_add_source
 -- Add source header to the request
 function add_source_header(txn)
   -- Don't log if map doesn't exist or sampled out
-  if (map == nil) or (sample_rate == 0) or (math.random() > sample_rate) then
-    txn.http:req_add_header('X-Smartstack-Origin', 'No-MAP-File')
+  if (map == nil) then
     return
   end
 
@@ -21,10 +20,10 @@ function add_source_header(txn)
   if ip == nil then
     ip = 'nil'
   end
+  
   src_svc = map:lookup(ip)
-
   if src_svc == nil then
-    src_svc = ip
+    return
   end
 
   -- Add header
--- a/src/synapse_tools/lua_scripts/add_source_header.lua
+++ b/src/synapse_tools/lua_scripts/add_source_header.lua
@@ -1,0 +1,33 @@
+-- Loads map into Lua script
+function init_add_source(txn)
+  -- Load map if not yet loaded
+  if map == nil then
+    local map_file = txn.f:env('map_file')
+    map = Map.new(map_file, Map.str)
+  end
+end
+core.register_action('init_add_source', {'tcp-req', 'http-req'}, init_add_source)
+
+-- Add source header to the request
+function add_source_header(txn)
+  -- Don't log if map doesn't exist or sampled out
+  if (map == nil) or (sample_rate == 0) or (math.random() > sample_rate) then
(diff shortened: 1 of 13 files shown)


src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -0,0 +1,33 @@
...
+    local map_file = txn.f:env('map_file')
+    map = Map.new(map_file, Map.str)
+  end
+end
+core.register_action('init_add_source', {'tcp-req', 'http-req'}, init_add_source)
+
+-- Add source header to the request
+function add_source_header(txn)
+  -- Don't log if map doesn't exist or sampled out
+  if (map == nil) or (sample_rate == 0) or (math.random() > sample_rate) then
```
ronin13: sample_rate needs to be removed.

src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -0,0 +1,33 @@
...
+    txn.http:req_add_header('X-Smartstack-Origin', 'No-MAP-File')
```
ronin13: No need to set header in this case

src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  -- Get source service
+  local ip = txn.f:src()
+  if ip == nil then
+    ip = 'nil'
+  end
+  src_svc = map:lookup(ip)
+
+  if src_svc == nil then
```
ronin13: Better to not add header in this case.

src/synapse_tools/lua_scripts/add_source_header.lua
avadhutp: Fixed

src/synapse_tools/lua_scripts/add_source_header.lua
avadhutp: Fixed; was an artifact of debugging.

src/synapse_tools/lua_scripts/add_source_header.lua
avadhutp: Fixed

avadhutp:
@jvperrin @matfra So, we tested this out on stagef (thanks to @ronin13) and it does generate the expected headers:
```
sudo tcpdump -nni eth0 -A port 31165 and 'tcp[13] & 8!=0'  | grep -E '^[a-zA-Z][[:alnum:]\-]+:' | grep X-Smartstack-Origin:
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on eth0, link-type EN10MB (Ethernet), capture size 262144 bytes
X-Smartstack-Origin: user--personalization.main.gita764bbf5.config04070982.6c1ed6f8-2f11-11e8-9a85-620da3a57bfa
X-Smartstack-Origin: ad--landscape.main.git082470fc.configf4e1eda3.bd09d77d-2f09-11e8-9a85-620da3a57bfa
X-Smartstack-Origin: ad--creative.main.git16a41955.config5bbbb2dc.6c301544-2f11-11e8-9688-e63b16ca4f39
```

avadhutp:
@matfra Nope, the file looks good as is :-)

PR #55, opened 2018-05-01
Pass a 0 as source for unidentifiable services

avadhutp:
Set `X-Smartstack-Origin` to `0` for unidentifiable services to avoid spoofing. This automatically overrides any `X-Smartstack-Origin` that the client may send.

### DIFF
--- a/dockerfiles/itest/itest/itest.py
+++ b/dockerfiles/itest/itest/itest.py
@@ -446,7 +446,7 @@ class TestGroupOne(object):
             url = 'http://localhost:%d%s' % (data['proxy_port'], data['healthcheck_uri'])
 
             # First, test with the service IP present in the map file
-            request = urllib2.Request(url=url)
+            request = urllib2.Request(url=url, headers={'X-Smartstack-Origin': 'Spoof-Value'})
             with contextlib.closing(
                 urllib2.urlopen(request, timeout=SOCKET_TIMEOUT)) as page:
                 assert page.info().dict['x-smartstack-origin'] == 'Test'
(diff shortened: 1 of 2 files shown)


dockerfiles/itest/itest/itest.py
```diff
@@ -172,296 +176,329 @@ def _sort_lists_in_dict(d):
...
+    def test_source_required_plugin_without_map_entry(self, setup):
+        # Test plugins with only HAProxy
+        if 'nginx' not in setup and 'both' not in setup:
+
+            name = 'service_two.main'
+            data = SERVICES[name]
+            url = 'http://localhost:%d%s' % (data['proxy_port'], data['healthcheck_uri'])
+
+            # First, test with the service IP present in the map file
+            request = urllib2.Request(url=url)
```
ronin13: Would be good to add a test on if the header is already present in the incoming request (and whether it is overwritten or not).

src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -23,7 +23,7 @@ function add_source_header(txn)
   
   src_svc = map:lookup(ip)
   if src_svc == nil then
-    return
+    src_svc = '0'
   end
 
   -- Add header
```
ronin13: As mentioned in testing section, would be good to verify if this overwrites an incoming request's header of same name (X-Smartstack-Origin).

ronin13: Also, as discussed, would be better to unset the header at the beginning (and test for it), so that even in the case of map lookup returning nil we don't pass any spoofed header to the destination.

PR #57, opened 2018-10-15
Make the lua script to add headers more performant and robust:

avadhutp:
- load the map right after the config is parsed
- avoid checking map's availability on each request
- reload the map, out of band, every 5s

### DIFF
--- a/src/synapse_tools/lua_scripts/add_source_header.lua
+++ b/src/synapse_tools/lua_scripts/add_source_header.lua
@@ -1,55 +1,24 @@
 map = {}
-map_file = ''
-
--- Get the path to the haproxy.cfg file
-function get_config_filename(cmdline)
-  local found = false
-  for s in string.gmatch(cmdline, '%g+') do
-    if s == '-f' then
-      found = true
-    elseif found then
-      return s
-    end
-  end
-end
-
--- Get the map_file path from haproxy.cfg's global stanza
-function get_map_file(config_file)
-   fh, err = io.open(config_file)
-   if err then
-     core.log(core.err, 'Cannot open the map file at ' .. config_file .. '. This will cause service authentication to fail!')
-   end
-
-  while true do
-     ln = fh:read()
-    if ln == nil then break end
-    if ln:find('setenv map_file') then
-       return ln:match('%S+/.*$')
-    end
-  end
-end
-
--- Get the cmdline used to run the haproxy process, because this cmdline
--- will also have the path to haproxy.cfg
-function get_cmdline()
-  local f = io.open('/proc/self/cmdline', "rb")
-  local cmdline = f:read("*all")
-  f:close()
-
-  return cmdline
-end
+map_file = nil
+refresh_interval = nil
 
 function refresh_map()
   map = Map.new(map_file, Map.str)
 end
 
+function init_vars()
+  map_file = os.getenv('map_file')
+  if map_file == nil then
+     core.log(core.err, 'map_file variable not set! This will cause authentication to fail for some requests.')
+   end
+
+  refresh_interval = os.getenv('map_refresh_interval')
+end
+
 -- This is run soon after haproxy parses haproxy.cfg. We will
 -- load the ip_to_svc map for the first time here.
 core.register_init(function()
-  local cmdline = get_cmdline()
-  local config_file = get_config_filename(cmdline)
-  map_file = get_map_file(config_file)
-
+  init_vars()
   refresh_map()
 end)
 
@@ -58,7 +27,7 @@ end)
 core.register_task(function()
   while true do
     refresh_map()
-    core.sleep(5)
+    core.sleep(refresh_interval)
   end
 end)
 
--- a/src/synapse_tools/lua_scripts/add_source_header.lua
+++ b/src/synapse_tools/lua_scripts/add_source_header.lua
@@ -1,12 +1,66 @@
--- Loads map into Lua script
-function init_add_source(txn)
-  -- Load map if not yet loaded
-  if map == nil then
-    local map_file = txn.f:env('map_file')
-    map = Map.new(map_file, Map.str)
+map = {}
+map_file = ''
+
+-- Get the path to the haproxy.cfg file
+function get_config_filename(cmdline)
+  local found = false
+  for s in string.gmatch(cmdline, '%g+') do
+    if s == '-f' then
+      found = true
+    elseif found then
+      return s
+    end
   end
 end
-core.register_action('init_add_source', {'tcp-req', 'http-req'}, init_add_source)
+
+-- Get the map_file path from haproxy.cfg's global stanza
+function get_map_file(config_file)
+   fh, err = io.open(config_file)
+   if err then
+     core.log(core.err, 'Cannot open the map file at ' .. config_file .. '. This will cause service authentication to fail!')
+   end
+
+  while true do
+     ln = fh:read()
+    if ln == nil then break end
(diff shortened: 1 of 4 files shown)


src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -1,12 +1,66 @@
...
+end)
+
+-- This will register a task, to refresh the ip_to_svc map,
+-- to run every 5 seconds
+core.register_task(function()
+  while true do
+    refresh_map()
+    core.sleep(5)
+  end
+end)
```
ronin13: The conditional below around 

   -- Don't log if map doesn't exist or sampled out

is not required right.

src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -1,12 +1,66 @@
...
+      found = true
+    elseif found then
+      return s
+    end
   end
 end
-core.register_action('init_add_source', {'tcp-req', 'http-req'}, init_add_source)
+
+-- Get the map_file path from haproxy.cfg's global stanza
+function get_map_file(config_file)
```
ronin13: This is not required. 

The setenv that haproxy does is through  http://man7.org/linux/man-pages/man3/setenv.3.html  which means it should be available to the lua script.

This was added in https://github.com/haproxy/haproxy/commit/1d54972789c00e504f6e396c8c2046d718fda26f

I tested with 

`  

-- Add source header to the request
function add_source_header(txn)
  -- First, explicitly remove any existing origin headers to avoid spoofing
  local log_text = 'AARDVARK ' .. os.getenv("map_file") .. '\n'
  txn.Info(txn, log_text)
  txn.http:req_del_header('X-Smartstack-Origin')


`

and in the scribe log I noticed

`

2018-10-16 02:44:51 dev8-uswest1cdevc informational AARDVARK /var/run/synapse/maps/ip_to_service.map.

`

src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -1,12 +1,66 @@
...
+
+  refresh_map()
+end)
+
+-- This will register a task, to refresh the ip_to_svc map,
+-- to run every 5 seconds
+core.register_task(function()
+  while true do
+    refresh_map()
+    core.sleep(5)
```
ronin13: Seems like this actually yields to haproxy scheduler (which is a good thing). Details https://github.com/arodland/haproxy/blob/master/doc/lua.txt#L450

src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -1,12 +1,66 @@
...
+function get_cmdline()
+  local f = io.open('/proc/self/cmdline', "rb")
+  local cmdline = f:read("*all")
+  f:close()
+
+  return cmdline
+end
+
+function refresh_map()
+  map = Map.new(map_file, Map.str)
```
ronin13: Not a blocker but need to check over a few min runs that there is no memory leak when re-initializing the map. Given that our haproxy uptime is not much, it is not a big concern but something we need to be wary of.

src/synapse_tools/lua_scripts/add_source_header.lua
ronin13: Also, can we add some logging around 

`
     src_svc = '0'

`

ie. on details of context and environment which can help us with debugging. Also, since this is rare, shouldn't impact.

src/synapse_tools/lua_scripts/add_source_header.lua
ronin13: Tested in init as well with:

`
core.register_init(function()
  local log_text = 'ARMADILLO ' .. os.getenv("map_file") .. '\n'
  core.log(2, log_text)
end)
`
and it  printed

`
[crit] 288/033824 (37867) : ARMADILLO /var/run/synapse/maps/ip_to_service.map.
`

src/synapse_tools/lua_scripts/add_source_header.lua
ronin13: Also, good to check (not a blocker) if '5' above can be made a parameter (that we can ideally pass through puppet).

src/synapse_tools/lua_scripts/add_source_header.lua
avadhutp: Done.

src/synapse_tools/lua_scripts/add_source_header.lua
avadhutp: There isn't much context we have when `src_svc == 0`.

src/synapse_tools/lua_scripts/add_source_header.lua
avadhutp: > Also, good to check (not a blocker) if '5' above can be made a parameter (that we can ideally pass through puppet).

Done. This now defaults to `5` but can be overridden via puppet.

src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -1,12 +1,35 @@
--- Loads map into Lua script
-function init_add_source(txn)
-  -- Load map if not yet loaded
-  if map == nil then
-    local map_file = txn.f:env('map_file')
-    map = Map.new(map_file, Map.str)
-  end
+map = {}
```
ronin13: We are checking for map == nil  below but that will never be true if {}

`
> {} == nil
false

`

I guess we need to change the check below to next(map) == nil since

`

> next({}) == nil
true

`

ronin13:
shipit otherwise.

ronin13:
`
 core.register_action('add_source_header', {'tcp-req', 'http-req'}, add_source_header)
`
Do we need for tcp-req as well? May be better to skip it for simplicity (and not catch healthchecks etc.)

src/synapse_tools/lua_scripts/add_source_header.lua
ronin13: As discussed, would be good to log the ip address itself along side any identifying header such as X-Operation-ID. (tmp_synapse_haproxy logging is fine for now).

src/synapse_tools/lua_scripts/add_source_header.lua
avadhutp: Just made a check for `map == {}`

avadhutp:
> `core.register_action('add_source_header', {'tcp-req', 'http-req'}, add_source_header)`
> Do we need for tcp-req as well? May be better to skip it for simplicity (and not catch healthchecks etc.)
> 
> I think we are also probably introspecting SYN/ACKs due to this.

Done.

src/synapse_tools/lua_scripts/add_source_header.lua
ronin13: @avadhutp  ^^^ needs can be added in a separate PR. (not a blocker here).

src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -1,20 +1,43 @@
...
+refresh_interval = nil
+
+function refresh_map()
+  map = Map.new(map_file, Map.str)
+end
+
+function init_vars()
+  map_file = os.getenv('map_file')
+  if map_file == nil then
+     core.log(core.err, 'map_file variable not set! This will cause authentication to fail for some requests.')
```
danielpops: Where does this log entry go? Can we alert on the presence or volume of these messages?

PR #58, opened 2018-10-22
Dragging synapse-tools into python 3

avadhutp:
- Updating synapse-tools to paasta-tools > 0.56.0 was essential as the build was failing because paasta-tools 0.56.0 pinned setuptools < 12; this did not allow the build to pass anymore
- paasta-tools 0.81.3 uses aiohttp, which is python 3 only; which means that synapse-tools would also have to be python 3 only
- we also removed the build for lucid as neither travis nor our internal Jenkins pipeline build on lucid anymore: this is because lucid does not even have py27

Marking:
@ronin13 as he has the most context behind this change
@mattmb because he performed a similar change for paasta-tools
@jnb as he is the original owner and continues to own this in core-services


avadhutp:
Merged into 0.13.22

PR #59, opened 2018-10-24
Make the add_source_header lua plugin more resilient to the absence of the map file

avadhutp:
I have added new integration test that, without the fixes I've made in this commit, failed with the same error that we saw in the last go (https://fluffy.yelpcorp.com/i/4Q14JJHcMtPCQgqKhx7tQ4CtsbH5rMG0.html):
```
[ALERT] 295/063605 (89911) : lua init: runtime error: ...packages/synapse_tools/lua_scripts/add_source_header.lua:6: 'new': failed to open pattern file </var/run/synapse/maps/ip_to_service.map>..
```

Now this test passes.

### DIFF
--- a/src/synapse_tools/lua_scripts/add_source_header.lua
+++ b/src/synapse_tools/lua_scripts/add_source_header.lua
@@ -2,6 +2,10 @@ map = {}
 map_file = nil
 refresh_interval = nil
 
+function log_error(err)
+  core.log(core.err, msg)
+end
+
 function refresh_map()
   map = Map.new(map_file, Map.str)
 end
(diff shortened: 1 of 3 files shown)


src/synapse_tools/lua_scripts/add_source_header.lua
```diff
@@ -2,14 +2,18 @@ map = {}
 map_file = nil
 refresh_interval = nil
 
+function log_error(err)
+  core.log(core.err, msg)
```
ronin13: Shouldn't this be err?  (the argument is err)

PR #64, opened 2018-12-04
Add typing annotations

vkhromov:


solarkennedy:
gotta fix/wait for the paasta-tools version, but I say ship if the tests pass

PR #65, opened 2019-01-08
Add error log option

matfra:
We run nginx without any kind of error logging. This could potentially
    hide issues. I'm investigating a case where requests are not making it
    to haproxy so enabling the logs is a first good step here.
    This change is a noop as I keep the same default values

I didn't bump pyyaml because the version that contains the fix for the security issue mentioned by github is not stable yet

See https://github.com/yaml/pyyaml/issues/243

### DIFF
--- a/src/synapse_tools/configure_synapse.py
+++ b/src/synapse_tools/configure_synapse.py
@@ -54,6 +54,8 @@ class NginxTopLevelConfig(TypedDict):
     restart_interval: int
     restart_jitter: float
     listen_address: str
+    log_error_target: str
+    log_error_level: str
 
 
 HAProxyTopLevelConfigExtraSections = TypedDict(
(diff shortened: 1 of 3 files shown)


src/synapse_tools/configure_synapse.py
```diff
@@ -54,6 +54,8 @@ class NginxTopLevelConfig(TypedDict):
     restart_interval: int
     restart_jitter: float
     listen_address: str
+    log_error_target: str
+    log_error_level: str
```
thebostik: I don't think these are needed here at all anymore. Since we aren't adding these as their own top-level dict keys, but we're instead adding the `error_log ...` string to the typed config dict under `contexts.main`.

vkhromov: Either add `nginx_` prefix here or remove the same prefix when you're accessing the config dictionary below.

vkhromov: This is exactly [what `mypy` saying](https://travis-ci.org/Yelp/synapse-tools/jobs/476620612):
```
synapse_tools/configure_synapse.py:230: error: Keys ('log_error_target', 'log_error_level') missing for TypedDict "NginxTopLevelConfig"
```

PR #67, opened 2019-04-22
Set x-smartstack-destination header instead of adding it to prevent casper proxy issues

kaisen:
Casper returns a 400 if duplicate X-Smartstack-Destination headers are found in the request. This can occur if envoy routes to haproxy which routes to casper (specifically for casper-enabled services).

This change updates the haproxy directive used to add the header so that the header is set (and replaced if it already exists) instead of being simply appended.

Both `make itest_trusty` and `make itest_xenial` pass, and I was able to confirm that `http-request set-header` works as intended with a locally running instance of haproxy. I also tested that the ACLs used still work with the new directive.


bplotnick:
lg2m Do we want to also do it for the backend (`X-Smartstack-Source`)?

kaisen:
Good idea, x-smartstack-source is now set using `http-request set-header` too.

PR #71, opened 2019-07-18
Build for bionic

jvperrin:
This is SMTSTK-228, but essentially I'm just adding the itests, dockerfiles, and general setup for building and testing this with bionic. None of the actual package code changes, just the testing infrastructure.

I also upgraded the nginx version used for trusty from `1.11.10` to `1.13.3` to match bionic (and the version we are using), and had to custom-built the nginx version in bionic to avoid an issue where the upstream nginx version uses `--with-stream=dynamic` but we use `--with-stream` meaning the config would be different between the itest and reality (and just for bionic too), so I thought building nginx in the container was the easiest and most consistent way out of that one.

### DIFF
--- a/dockerfiles/bionic/Dockerfile
+++ b/dockerfiles/bionic/Dockerfile
@@ -1,5 +1,4 @@
 FROM ubuntu:bionic
-MAINTAINER John Billings <billings@yelp.com>
 
 RUN apt-get update && apt-get -y install \
     debhelper \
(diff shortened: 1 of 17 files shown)


dockerfiles/bionic/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ubuntu:bionic
+MAINTAINER John Billings <billings@yelp.com>
```
bplotnick: should we change this 😄

jvperrin: Haha, yeah probably. I think just removing it is best, since it's deprecated anyway (https://docs.docker.com/engine/reference/builder/#maintainer-deprecated)

PR #72, opened 2019-09-18
Reload nginx really gracefully

vkhromov:
Introduce `nginx_reload_script` configuration option, allowing to
specify a script which will be used to reload `nginx`.

Change the existing actions in such a way that while reloading `nginx`,
after we sent `SIGUSR2` to `nginx` asking it to launch a new instance of
itself, let's check that the new instance actually launched before
killing the old one.

### DIFF
--- a/src/synapse_tools/configure_synapse.py
+++ b/src/synapse_tools/configure_synapse.py
@@ -186,13 +186,13 @@ def set_defaults(
         ('nginx_prefix', '/var/run/synapse/nginx_temp'),
         ('nginx_config_path', '/var/run/synapse/nginx.cfg'),
         ('nginx_pid_file_path', '/var/run/synapse/nginx.pid'),
+        ('nginx_reload_script',
+            r"""/bin/bash -c 'set -ue -o pipefail; q() { pidfile=$1; oldpid=$(cat $pidfile); kill -USR2 $oldpid; sleep 2; newpid=$(cat $pidfile); if [ $oldpid -eq $newpid ]; then return 1; fi; kill -WINCH $(cat $pidfile.oldbin); kill -QUIT $(cat $pidfile.oldbin); }; q $0'"""),
         ('nginx_proxy_proto', False),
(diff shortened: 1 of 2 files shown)


src/synapse_tools/configure_synapse.py
```diff
@@ -186,13 +186,32 @@ def set_defaults(
...
         ('nginx_pid_file_path', '/var/run/synapse/nginx.pid'),
+        ('nginx_reload_script',
+            r"""
+            /bin/bash -c '
+            set -ue -o pipefail
+            q() {
+                pidfile=$1
+                oldpid=$(cat $pidfile)
+                kill -USR2 $oldpid
+                sleep 2
```
keymone: this looks like something that can cause mysterious bugs. don't know what to suggest though.

src/synapse_tools/configure_synapse.py
```diff
@@ -186,13 +186,32 @@ def set_defaults(
...
+                newpid=$(cat $pidfile)
+                if [ $oldpid = $newpid ]; then
+                    # the new pid is the same as the old pid, which means that
+                    # nginx failed to launch the new instance of itself.
+                    return 1
+                fi
+                kill -WINCH $oldpid
```
mattmb: maybe we should still stick with the "oldbin" file? I agree the oldpid is the right ID but maybe there's some other checks that nginx has performed so that if that file exists there's a better likelihood that the old process is safe to kill?

src/synapse_tools/configure_synapse.py
mattmb: heh, totally. I mean the good news is we're not adding this sleep now. We're just adding a check after the sleep to try and assert the new nginx hasn't died before we kill the old one. But yes, there's still a chance of a race here if we are unlucky.

vkhromov:
To test it I made some changes in a service port to make sure that the command still gracefully reload `nginx`.
I also run the command to make sure that it fails if the new nginx pid is the same as the old one.

PR #73, opened 2019-09-23
Remove trusty from tests

vkhromov:
We don't use Trusty, so let's remove it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 env:
-  - MAKE_TARGET=itest_trusty
   - MAKE_TARGET=itest_xenial
   - MAKE_TARGET=itest_bionic
   - MAKE_TARGET=mypy


solarkennedy:
We still have 52 servers at yelp that run trusty and synapse per `mco facts lsbdistrelease -C synapse_simple::install`

vkhromov:
@solarkennedy, `itest-trusty` suddenly stopped working for the master, e.g. see https://travis-ci.org/Yelp/synapse-tools/jobs/560794207, are you sure that that really worth fixing?

solarkennedy:
We are going to have to live with these trusty boxes for a while, migrating them is outside of our control. Yes, we need to fix it because if we don't, we won't be able to upgrade synapse on these servers.

vkhromov:
@solarkennedy can you or somebody else from SF please fix that, so I could merge my DAR-related fix #72 tomorrow?

PR #75, opened 2019-10-02
add service to ip maapping for k8s services

EmanekaT:
I only tested with make itest. Most of the code is copied from hpa-metrics-collector which works. Instead of checking if the instance is a k8s node, I just catch the error if it is not.

### DIFF
--- a/src/synapse_tools/generate_container_ip_map.py
+++ b/src/synapse_tools/generate_container_ip_map.py
@@ -2,6 +2,7 @@
 import argparse
 import os
 import socket
+import requests
 from typing import Iterable
 from typing import Optional
 from typing import Mapping
@@ -54,6 +55,25 @@ def extract_taskid_and_ip(
                 # For compatibility with MESOS_TASK_ID format.
                 task_id = task_id.replace('_', '--')
                 service_ips_and_ids.append((ip_addr, task_id))
+
+    # For kubernetes. If the ec2 instance is not a k8s instance, this will throw an error and be skipped.
+    # Get all pod metadata on this node
+    try:
+        node_info = requests.get(f"http://169.254.255.2 54:10255/pods/").json()
+        for pod in node_info['items']:
+            # Fiter out instances that are not running
+            if pod['status']['phase'] != 'Running' \
+                    or 'smartstack_registrations' not in pod['metadata'].get('annotations', {}):
+                continue
+            labels = pod['metadata']['labels']
+            service = labels['yelp.com/paasta_service']
+            instance = labels['yelp.com/paasta_instance']
+            task_id = f'{service}.{instance}'.replace('_', '--')
+            pod_ip = pod['status']['podIP']
+            service_ips_and_ids.append((pod_ip, task_id))
+    except Exception:
+        pass
+
     return service_ips_and_ids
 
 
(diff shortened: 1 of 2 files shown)


src/synapse_tools/generate_container_ip_map.py
```diff
@@ -54,6 +55,25 @@ def extract_taskid_and_ip(
                 # For compatibility with MESOS_TASK_ID format.
                 task_id = task_id.replace('_', '--')
                 service_ips_and_ids.append((ip_addr, task_id))
+
+    # For kubernetes. If the ec2 instance is not a k8s instance, this will throw an error and be skipped.
+    # Get all pod metadata on this node
+    try:
+        node_info = requests.get(f"http://169.254.255.2 54:10255/pods/").json()
```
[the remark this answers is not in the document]
EmanekaT: @qui  I checked synapse run as root. I'm confident it can read kubelet api since I tested with hpa-metrics-collector before.

src/synapse_tools/generate_container_ip_map.py
```diff
@@ -54,6 +55,23 @@ def extract_taskid_and_ip(
...
+    except Exception:
+        return service_ips_and_ids
+
+    for pod in node_info['items']:
+        labels = pod['metadata']['labels']
+        if 'yelp.com/paasta_service' in labels and 'yelp.com/paasta_instance' in labels:
+            service = labels['yelp.com/paasta_service']
+            instance = labels['yelp.com/paasta_instance']
+            task_id = f'{service}.{instance}'.replace('_', '--')
```
chlgit: Could you make a function and use that function here as well as line 51-56?

src/synapse_tools/generate_container_ip_map.py
```diff
@@ -54,6 +55,23 @@ def extract_taskid_and_ip(
...
+            pod_ip = pod['status']['podIP']
```
chlgit: Are you sure it's 'podIP' not 'pod_ip'?

EmanekaT: Yes. I ran this in hpa-metrics-collector before. And it's Go convention, not python. (I love camelcase :))

src/synapse_tools/generate_container_ip_map.py
EmanekaT: hmm, I don't really want to make a seperate function unless the code is reused or really long. BTW do you know where to add tests? I don't find any tests for generate_container_ip_map.py file. I don't want to add a new test for that file.

chlgit: I feel line 51-56 and line 69 - 71 should be enough to make a function so that we would have a single source to make sure tast_id follows the format. In terms of testing, I was thinking to add the test here https://github.com/Yelp/synapse-tools/blob/master/dockerfiles/itest/itest/itest.py#L210.

src/synapse_tools/generate_container_ip_map.py
chlgit: Cool, thanks for the confirmation.

EmanekaT:
I tested it on a k8s node and it works.
https://fluffy.yelpcorp.com/i/hF6xcTQHTCjrq1x4wV7zBxgmmGq54px8.html

PR #76, opened 2019-10-18
Add per endpoint timeout support

matthewbentley:
The purpose of this is to allow service owners to add (for example):
```
endpoint_timeouts:
  "/example/endpoint": 10000
  "/other/endpoint": 15000
```
to their config, to override the server timeouts for just those endpoints.

This seemed straightforward at first, but is significantly complicated by discover/advertise options.  Currently, each advertise option creates a new backend.  The initial plan for this was to create a new backend for each path with an overridden timeout.  The result is that we have to create new backends for the cartesian product of (advertise_types, path_timeout_overrides).

The tarpit options don't work for the per-endpoint-timeouts backends, but they also don't work for any advertise backend other than the default.

I did a manual end-to-end check that the endpoint matching actually works and applies the proper timeouts.

I'll bump the version to v0.14.0 after merging to master.


src/synapse_tools/configure_synapse.py
```diff
@@ -456,10 +463,16 @@ def get_backend_name(
...
+    if endpoint_name != "default":
+        endpoint_ext = f".{endpoint_name}_timeouts"
+    else:
+        endpoint_ext = ""
     if advertise_type == discover_type:
-        return service_name
+        return service_name + endpoint_ext
     else:
+        return f"{service_name}.{advertise_type}{endpoint_ext}"
         return '%s.%s' % (service_name, advertise_type)
```
jvperrin: Unreachable return?

matthewbentley: Whoops yep very WIP.

src/synapse_tools/configure_synapse.py
```diff
@@ -495,17 +519,22 @@ def generate_acls_for_service(
             service_name=service_name,
             discover_type=discover_type,
             advertise_type=advertise_type,
+            endpoint_name=endpoint_name,
         )
 
+        if endpoint_name != 'default':
+            path = endpoint_timeouts[endpoint_name]['endpoint']
+            path_acl_name = f' {backend_identifier}_path'
```
ddelnano: nit: this is excess leading whitespace isn't it?

dockerfiles/itest/itest/itest.py
```diff
@@ -221,6 +238,10 @@ def test_synapse_services(self, setup):
             'service_three_chaos.main',
             'service_two.main',
             'service_three.logging',
+            'service_three_endpoint_timeout.main.foo_bar_timeouts',
+            'service_three_endpoint_timeout.main',
+            'service_three_endpoint_timeout.main.region.foo_bar_timeouts',
+            'service_three_endpoint_timeout.main.region',
```
ddelnano: Our itest coverage just proves synapse starts up correctly but doesn't necessarily validate that our haproxy configuration works right? Have you put any though into how we could test that if so?

src/synapse_tools/configure_synapse.py
matthewbentley: that's intentional, so line 537 doesn't need a space

dockerfiles/itest/itest/itest.py
matthewbentley: That's what we need to figure out next

matthewbentley: I added a basic test that it can route to /foo/bar, but I still need to figure out if it's actually being routed properly

matthewbentley:
This is mostly ready for review.  I can probably add some more comments or tests if that would be helpful.

src/synapse_tools/configure_synapse.py
```diff
@@ -479,14 +498,37 @@ def _get_socket_path(
...
+    endpoint_timeouts: Dict[str, EndpointTimeoutConfig],
+    per_endpoint_timeouts_enabled: bool,
+) -> Iterable[Tuple[str, str]]:
+    """Get the cartesian product of advertise types and endpoint timeout overrides.
+    This is used to make the list of backends for synapse.conf.json.
+    """
+    if per_endpoint_timeouts_enabled:
+        endpoint_timeouts_names = list(endpoint_timeouts.keys()) + ['default']
+    else:
+        endpoint_timeouts_names = ['default']
```
sergsyelp: I think we may end up in a situation when there is a valid endpoint named 'default' so we need some kind of constant here that would not interfere with legitimate endpoint names.

dockerfiles/itest/itest/itest.py
```diff
@@ -29,6 +29,23 @@
...
+        'ip_address': socket.gethostbyname(CONTAINER_PREFIX + 'servicethree_endpoint_timeout_1'),
+        'port': 1024,
+        'proxy_port': 20070,
+        'mode': 'http',
+        'healthcheck_uri': '/my_healthcheck_endpoint',
+        'discover': 'habitat',
+        'advertise': ['habitat', 'region'],
+        'endpoint_timeouts': {
+            'example_endpoint': {
+                'endpoint': '/foo/bar',
```
solarkennedy: Why does this need to be listed twice? Why not just have this be the key to the dictionary?

matthewbentley: The original reason for having the dict key be different is that we need something to use as the haproxy backend names (where `/` is not a valid character).  We should replace that `/` with `.` or `_` or something if that would make the schema easier to understand.

src/synapse_tools/configure_synapse.py
matthewbentley: I fixed this by adding a `sentinel` object that is used in place of the literal string `'default'`

dockerfiles/itest/itest/itest.py
matthewbentley: I have a solution for this, I'll push it as soon as the itests pass.

src/synapse_tools/configure_synapse.py
```diff
@@ -26,13 +28,19 @@
...
 from synapse_tools.config_plugins.base import ServiceInfo
 from synapse_tools.config_plugins.base import SynapseToolsConfig
 from synapse_tools.config_plugins.registry import PLUGIN_REGISTRY
 from synapse_tools.haproxy_synapse_reaper import DEFAULT_REAP_AGE_S
 from yaml import CLoader  # type: ignore
 
 
+# This is to keep track of the "default" haproxy section
+#  (eg no overridden timeout)
+HAPROXY_DEFAULT_SECTION_SENTINAL = object()
```
matthewbentley: The use of this object was what I am most apprehensive about; on one hand I feel like it's a bit of a hack to paper over the fact that the default frontend-backend pair is different from the rest of the backends, but on the other hand it would be a significant refactor of the code to fix that.

dockerfiles/itest/itest/itest.py
matthewbentley: I updated this to replace `/` with `__`.  This would break in the (unlikely) case where a service has (eg) `/foo/bar` and `/foo__bar` as endpoints, which I don't think we need to worry about.

src/synapse_tools/configure_synapse.py
```diff
@@ -26,13 +28,19 @@
...
 from synapse_tools.config_plugins.base import ServiceInfo
 from synapse_tools.config_plugins.base import SynapseToolsConfig
 from synapse_tools.config_plugins.registry import PLUGIN_REGISTRY
 from synapse_tools.haproxy_synapse_reaper import DEFAULT_REAP_AGE_S
 from yaml import CLoader  # type: ignore
 
 
+# This is to keep track of the "default" haproxy section
+#  (eg no overridden timeout)
+HAPROXY_DEFAULT_SECTION_SENTINEL = object()
```
solarkennedy: I'm still looking at this, but I've never seen this pattern before can I don't really see how it gets set...

matthewbentley: Originally [here](https://github.com/Yelp/synapse-tools/pull/76/files#diff-c9826d5a06e0f430bb006b1a732896f8R516) I just added the string `'default'` to the list, to keep track of the "default" frontend/backend that needs to be added.  But any string used there could break (if there was an endpoint with that name), so using a new `object()` gets around that (it's only compared for in a few places).

This is the part that I knew would be unclear, but wasn't sure of a way around it without re-writing large parts of this program.

matthewbentley: Looking at this again, I think it would be safe to use a string here, since endpoint names _must_ start with a `/` now that I updated the code to not use the nested dictionary, if that would be more clear.

solarkennedy: Idk, something anything. 

Just assure me it is going to be treated as a constant (like a string) and not some sort of global object variable

matthewbentley: I changed this back to just be the string "default", and also made it `Final` with mypy.

solarkennedy:
This is a significant change to how we make backends, and it will "explode" the backend names with lots of timeouts, etc.

I'm mostly concerned about the "load" on haproxy and the generating infrastructure in general.

It also looks like this is going to be backwards and fowards compatible, because it doesn't change anything in zookeeper. Great! (right)

Ship using the standard hiera-esq deployment strategy.

PR #78, opened 2019-12-02
Special case per-endpoint timeouts for "/" to do full string match

matthewbentley:
There is no reason to do a prefix match for "/", so special case it for now

We might decide to just switch to regex matching later, but this enables users who need to override "/" to do so for now

Also update a unit test


joelesalas:
Looks good to me!

ddelnano:
🚢

PR #80, opened 2020-04-23
Use new label scheme in generate_container_ip_map

keymone:
Forgot to update this script when I migrated things last time.

### DIFF
--- a/src/synapse_tools/generate_container_ip_map.py
+++ b/src/synapse_tools/generate_container_ip_map.py
@@ -65,14 +65,8 @@ def extract_taskid_and_ip(
 
     for pod in node_info['items']:
         labels = pod['metadata']['labels']
-        service = labels.get(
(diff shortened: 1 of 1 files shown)


src/synapse_tools/generate_container_ip_map.py
```diff
@@ -65,11 +65,17 @@ def extract_taskid_and_ip(
 
     for pod in node_info['items']:
         labels = pod['metadata']['labels']
-        if 'yelp.com/paasta_service' in labels and \
-           'yelp.com/paasta_instance' in labels and \
+        service = labels.get(
```
solarkennedy, line 68: By now isn't it safe to assume we are using the new scheme everywhere?

keymone: i suppose.. i'm just following my internal deprecation protocols. i'll update.

PR #85, opened 2020-08-06
Ensure internal yelp builds use internal pypi

mattmb:
Mostly for security reasons we don't want to build from public pypi
internally. We do want to use it for travis of course. Had to update
docker compose to make it pass a build arg. Copy paste of approach in
https://github.com/Yelp/nerve-tools/pull/76

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 env:
-  - MAKE_TARGET=itest_xenial
-  - MAKE_TARGET=itest_bionic
-  - MAKE_TARGET=mypy
+  global:
+    - PIP_INDEX_URL=https://pypi.python.org/simple
+  jobs:
+    - MAKE_TARGET=itest_xenial
+    - MAKE_TARGET=itest_bionic
+    - MAKE_TARGET=mypy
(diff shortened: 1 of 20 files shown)


mattmb:
The build is broken on master because Travis have helpfully blocked the IP range that haproxy.org lives in: https://travis-ci.community/t/access-to-external-git-repo-not-possible-anymore/8975

I'll probably ditch that last commit since it doesn't help now I know it's blocked at IP level. Anyway, otherwise I think this is good to go, here's the fpbs of the internal builds:
https://fluffy.yelpcorp.com/i/ZWkHFwF0L8l6WNdXHq2Dc6LWW5lD89ND.html
https://fluffy.yelpcorp.com/i/9KMF3VSZ7vF2V44fLXJrCn0tGSWBjbpQ.html

vkhromov:
```
Step 2/7 : ARG PIP_INDEX_URL=https://pypi.yelpcorp.com/simple
 ---> Using cache
```
I know that it probably cached a right version, but to be 100% sure could you please remove all local Docker artefacts before running tests?

PR #86, opened 2020-08-06
Add envoy migration toggles to allow disabling haproxy on a per-namespace basis

chriskuehl:
Internal ticket: MESH-914

This PR teaches configure_synapse how to respect a `envoy_migration.yaml` config file, which has a format like this:

```yaml
migration_enabled: {true|false}
namespaces:
    {namespace}:
        state: {synapse|dual|envoy}
```

When a namespace is unconfigured or in `synapse` or `dual` state, synapse will configure an haproxy listener on that port just as it usually does. When a namespace is in `envoy` state, no listener will be configured on that port, but synapse will continue to populate the discovery JSON files (the files under `/run/synapse/services`).

## Testing Done

* Existing tests pass
* Added some new tests
* Manually installed and tested on a Yelp devbox; internal users can inspect a diff of the synapse config [here](https://fluffy.yelpcorp.com/i/XshJLKrb3qNSs2rNbCFrG9pFMlnfbSSJ.html) (`example_happyhour.main` and `compute-infra-test-service.main_tcp` were both configured in Envoy-only mode for this test). I also confirmed that synapse stopped listening on those ports while continuing to populate the JSON discovery files.

### DIFF
--- a/src/synapse_tools/configure_synapse.py
+++ b/src/synapse_tools/configure_synapse.py
@@ -65,6 +65,15 @@ class NginxTopLevelConfig(TypedDict):
     listen_address: str
 
 
+class EnvoyMigrationNamespaceConfig(TypedDict):
+    state: str
+
+
+class EnvoyMigrationConfig(TypedDict):
+    migration_enabled: bool
+    namespaces: Mapping[str, EnvoyMigrationNamespaceConfig]
+
+
 HAProxyTopLevelConfigExtraSections = TypedDict(
     'HAProxyTopLevelConfigExtraSections',
     {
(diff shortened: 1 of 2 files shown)


src/synapse_tools/configure_synapse.py
```diff
@@ -1049,6 +1068,11 @@ def get_my_grouping(grouping_type: str) -> str:
         return fd.read().strip()
 
 
+def get_envoy_migration_config(path: str) -> EnvoyMigrationConfig:
+    with open(path) as f:
```
vkhromov, line 1072: Please note that the file could be absent on a fresh hosts, we need to make sure that Envoy will crash-loop or something like that and not stuck until the file will be created.

chriskuehl: configure_synapse runs as a cron job once per minute so it should keep trying until the file exists. Do you think it's worth handling the case where the file doesn't exist? I can add that logic in, but also don't want to mask real errors (i.e. where the file is missing for a long time incorrectly).

vkhromov: I think crashing is fine, it just need to be able to properly recover after that.

PR #88, opened 2020-08-13
Only disable the nginx listener when a namespace is Envoy-only

chriskuehl:
Internal ticket: MESH-931

The current behavior is to disable both the haproxy and nginx listener when a namespace is Envoy-only. There's a race condition when disabling a namespace where haproxy (the service that listens on a UNIX socket) reloads before nginx (the service that listens on the proxy port and proxies to haproxy) which results in 503s being served for about 10 seconds:

```
$ curl localhost:20139
<html>[...]

$ curl localhost:20139
503 Service Unavailable
Served by haproxy-synapse on 10-65-79-67-useast1aprod.useast1-prod.yelpcorp.com.

$ curl localhost:20139
503 Service Unavailable
Served by haproxy-synapse on 10-65-79-67-useast1aprod.useast1-prod.yelpcorp.com.

$ curl localhost:20139
curl: (7) Failed to connect to localhost port 20139: Connection refused
```

This is bad because if nginx and envoy are sharing the port, we may serve 503s from nginx for a few seconds depending on how the traffic is split.

The new approach is to just keep the haproxy listener around and only tear down the nginx listener. This avoids complex coordination that would be required to ensure they get torn down / brought up in the correct order to avoid serving errors, and should be effectively equivalent since nothing will be using the haproxy UNIX socket.

## Verification

Here's what a diff to the synapse config looks like now when disabling a namespace:

```diff
--- before      2020-08-13 09:59:53.946522493 -0700
+++ after       2020-08-13 10:05:27.145168974 -0700
@@ -30346,32 +30320,6 @@
             },
             "use_previous_backends": false
         },
-        "example_happyhour.envoy.nginx_listener": {
-            "default_servers": [
-                {
-                    "host": "unix",
-                    "port": "/var/run/synapse/sockets/example_happyhour.envoy.prxy"
-                }
-            ],
-            "discovery": {
-                "method": "base"
-            },
-            "file_output": {
-                "disabled": true
-            },
-            "haproxy": {
-                "disabled": true
-            },
-            "nginx": {
-                "mode": "tcp",
-                "port": 20139,
-                "server": [
-                    "proxy_timeout 3610s",
-                    "proxy_protocol on"
-                ]
-            },
-            "use_previous_backends": true
-        },
         "example_happyhour.main": {
             "default_servers": [],
             "discovery": {
```

I manually tested a built package when toggling off a namespace and it cleanly goes from working to connection refused:
```
$ while :; do curl -sSI localhost:20139/status | head -n1; sleep 0.05; done
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 200 OK
curl: (7) Failed to connect to localhost port 20139: Connection refused
curl: (7) Failed to connect to localhost port 20139: Connection refused
curl: (7) Failed to connect to localhost port 20139: Connection refused
curl: (7) Failed to connect to localhost port 20139: Connection refused
curl: (7) Failed to connect to localhost port 20139: Connection refused
curl: (7) Failed to connect to localhost port 20139: Connection refused
curl: (7) Failed to connect to localhost port 20139: Connection refused
```

### DIFF
--- a/src/synapse_tools/configure_synapse.py
+++ b/src/synapse_tools/configure_synapse.py
@@ -598,14 +598,10 @@ def generate_configuration(
         if proxy_port is not None and proxy_port < 0:
             continue
 
-        # When the migration to Envoy is enabled and this namespace is marked
-        # as Envoy-only, skip setting up load balancing (but keep the discovery
-        # files).
-        if (
+        namespace_is_envoy_only = (
             envoy_migration_config['migration_enabled'] and
             envoy_migration_config['namespaces'].get(service_name, {'state': 'synapse'}).get('state') == 'envoy'
-        ):
-            proxy_port = None
+        )
 
         # Note that at this point proxy_port can be:
         # * valid number: Wants Load balancing (HAProxy/Nginx)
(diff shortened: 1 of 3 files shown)


chriskuehl:
@ymilki yup, but I think they'll be taken care of as part of the cleanup we would need to do anyway when finally removing synapse—it shouldn't require an extra step.

src/synapse_tools/configure_synapse.py
```diff
@@ -703,18 +699,23 @@ def generate_configuration(
...
-                synapse_config['services'][listener_name] = (
-                    _generate_nginx_for_watcher(
-                        service_name=service_name,
-                        service_info=cast(ServiceInfo, service_info),
-                        synapse_tools_config=synapse_tools_config,
+            # When the migration to Envoy is enabled and this namespace is marked
+            # as Envoy-only, we still provide discovery via JSON files and set up
+            # the haproxy backend (to avoid a race condition in MESH-931), but
+            # don't configure an nginx listener on the proxy port.
+            if namespace_is_envoy_only is False:
```
vkhromov, line 706:
```
if not namespace_is_envoy_only:
```
?

chriskuehl: Heh, it's a bit of a quirk of mine that I tend to prefer explicit boolean comparisons as I think they're more readable, especially in the negative case ("is the boolean false?" reads better to me than "is not the boolean?"). I know this isn't really mainstream, though; I'll change this one and ship.

PR #90, opened 2020-08-26
Add reuseport_enable toggle option for nginx listeners

kaisen:
Internal ticket: MESH-862

Using the same config namespace as #86, this PR adds a `reuseport_enable` option that will configure all nginx listeners to use the SO_REUSEPORT socket option.

```yaml
migration_enabled: {true|false}
reuseport_enabled: {true|false}
namespaces:
    {namespace}:
        state: {synapse|dual|envoy}
```

The new option respects the existing `migration_enable` option as a "master kill switch", meaning if `migration_enable` is False, then SO_REUSEPORT will never be used.

## Testing Done

* Existing tests pass
* Added assert statements to verify the existence of the `reuseport` option in nginx listeners
* Added tests to cover the permutations of `migration_enabled` and `reuseport_enabled`

### DIFF
--- a/dockerfiles/itest/itest/srv-configs/service_mesh/envoy_migration.yaml
+++ b/dockerfiles/itest/itest/srv-configs/service_mesh/envoy_migration.yaml
@@ -1,2 +1,3 @@
 migration_enabled: false
+reuseport_enabled: false
 namespaces: {}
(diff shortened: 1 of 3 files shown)


vkhromov (approved):
:ship:

kaisen:
> lgtm!
> 
> Do we know what happens when we switch nginx from listening without SO_REUSEPORT to using it? I'm curious if we'll see a brief disruption in traffic or if nginx can change that without losing traffic somehow.

I haven't found anything online that suggests otherwise, since nginx has been fine doing reloads with the existing config. This will be good to verify when we move to kubestage.

chriskuehl:
I saw that @Rob-Johnson commented on MESH-900 and said that nginx seemed to handle it without downtime, so hopefully it works out fine.

PR #97, opened 2021-03-03
add more logging and timing details for map script

Rob-Johnson:
testing the build on a k8s host

```
robj@10-40-11-56-uswest1adevc:~  % sudo /usr/bin/generate_container_ip_map --k8s
/opt/venvs/synapse-tools/lib/python3.6/site-packages/paasta_tools/utils.py:1442: UserWarning: clog is unavailable
  warnings.warn("clog is unavailable")
querying kubelet for pods
completed query for pods in 9.355ms
writing new map file content
```

really what I'm trying to get here are
1) early failure with a timeout of 2s (I don't think it should ever take this long)
2) logging of what a normal response time is


src/synapse_tools/generate_container_ip_map.py
```diff
@@ -63,7 +63,11 @@ def extract_taskid_and_ip_mesos(
 def extract_taskid_and_ip_k8s() -> Iterable[Tuple[str, str]]:
     service_ips_and_ids = []
 
-    node_info = requests.get("http://169.254.255.254:10255/pods/").json()
+    print('querying kubelet for pods')
+    response = requests.get("http://169.254.255.254:10255/pods/", timeout=10)
```
ronin13, line 67: Not blocking - Since we run this generator every 5 seconds, may be good to fail fast here, so something < 5s? (may be 2 or 3?)

ronin13 (approved):

src/synapse_tools/generate_container_ip_map.py
ronin13: This is based on assumption that we want the file updated every 5s and not be stale for longer periods.

ronin13 (approved):

danielpops:
Not sure if you want the scope creep, but it would be nice to also address https://jira.yelpcorp.com/browse/COMPINFRA-486 while you're here :) to get rid of the `clog is unavailable` noise in the logs

danielpops (approved):

src/synapse_tools/generate_container_ip_map.py
Rob-Johnson: agreed - reduced timeout to 2 seconds

danielpops (approved):
FYI it doesn't look like there's any github action set up on this repo to upload to pypi

Rob-Johnson:
> FYI it doesn't look like there's any github action set up on this repo to upload to pypi

yeah, you're right, but I'm not sure if there is anything actually using it as a python package?

Rob-Johnson:
> > FYI it doesn't look like there's any github action set up on this repo to upload to pypi
> 
> yeah, you're right, but I'm not sure if there is anything actually using it as a python package?

I think we used to upload deb packages to bin tray, which we no longer have available

PR #100, opened 2021-08-04
migrate to GitHub actions and fix the build

Rob-Johnson:
1) migrate to github actions and remove travis references
2) fix the build
3) try and remove all traces of trusty, and remove pins of things that no longer need to be pinned


keymone (approved):
oh boy, that must have been painful

PR #103, opened 2021-09-29
Make generate_container_ip_map only look at non-terminal Pods

nemacysts:
We currently filter out failed Pods, but Tron and other workloads at
Yelp are now launching Pods that end up in other non-Failed terminal
states such as Succeeded.

This change makes it so that we only consider Pods that are actually
active (i.e., running something).


nemacysts:
to test: I packaged this as a deb and ran generate_container_ip_map on a box that had run some Tron jobs

original (includes Pods in terminal states): https://fluffy.yelpcorp.com/i/1Sb88002TRblrZf1VbbGqTx0jwSc4qGV.html#L7-8,L11,L13,L23,L27,L29,L31,L37,L42,L49
this PR (does not include Pods in terminal states): https://fluffy.yelpcorp.com/i/619hTxgClPkl4g2sTxXBjXS6fGXr3cfm.html

danielpops (approved):

nemacysts:
note: I based this on https://github.com/Yelp/synapse-tools/pull/101/files

src/synapse_tools/generate_container_ip_map.py
```diff
@@ -119,10 +119,14 @@ def extract_taskid_and_ip_k8s() -> Iterable[Tuple[str, str]]:
...
-            if service is not None and \
-               instance is not None and \
-               status != 'Failed' and \
-               'podIP' in pod['status']:
+            if (
+                service is not None and
+                instance is not None and
+                # only look at Pods that are actually running since IPs might be reused
+                # after a Pod has completed
+                status == "Running" and
```
danielpops, line 127: Reading up on https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase i see there are `phase` and `state` identifiers. Can we do some analysis on whether we need to include others such as `phase=Pending` or `state=Waiting`?

danielpops: @nemacysts also, given my new understanding that there is a difference between `phase` and `state`, can we rename this variable accordingly? :)

nemacysts: Only Pods have a `Phase` attribute and only containers inside said Pods have a `State` attribute - so we should be ok since (unless I'm missing something) we only want to inspect Pods, but I can go ahead and rename this to `pod_phase` for clarity

I'm actually not sure at what point Pods are actually given an IP address (i.e., is a Pod IP assigned while a Pod is still Pending and not yet assigned to a kubelet, is it instead when a kubelet has been assigned, or is it when the kubelet is actually setting up containers) - I was hoping that robj or vkhromov could chime in here :p

That said, I *think* it'd be safe to include `Pending` since if there's no `podIP` in the Pod status dict then we'd skip processing that Pod. I'll add this in a new commit and worst case  robj or vkhromov can revert that commit if they want to merge this in EU time and don't approve :)

danielpops: ack, thanks for digging. that sounds good to me

danielpops (approved):

Rob-Johnson (approved):

Rob-Johnson (approved):

PR #107, opened 2021-11-03
Use namespace_denylist option to filter out namespaces

keymone:
a followup to https://github.com/Yelp/synapse-tools/pull/106, filtering out a few noisy namespaces is less error-prone than having to keep a list of allowed ones.


vkhromov:
> ugh.. i've somewhat surrendered, there were no changes to how mypy tests are run, they pass in my environment, can't figure out why is mypy analyzing stuff in .tox/.../site-packages folders o_0

Not sure why it blew up at this specific PR, but we create `.tox` inside `synapse-tools` subdirectory:
```
mypy create: /home/runner/work/synapse-tools/synapse-tools/src/.tox/mypy
```
and later ask `mypy` to analyze `synapse_tools` with all its subdirectories including `.tox`:
```
mypy runtests: commands[0] | mypy synapse_tools
```

Could you try using `--exclude` (https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-exclude to ignore `.tox` dir?

keymone:
ok, bumping mypy was enough in itself (version we were using didn't have --exclude option)

src/synapse_tools/configure_synapse.py
```diff
@@ -1013,18 +1013,33 @@ def get_my_grouping(grouping_type: str) -> str:
...
     with open(path) as f:
-        return yaml.load(f, Loader=yaml.CSafeLoader)  # type: ignore
+        return yaml.load(f, Loader=yaml.CSafeLoader)
 
 
 def get_allowed_namespaces(
+    namespaces: Sequence[Tuple[str, ServiceNamespaceConfig]],
     allowlist: Optional[Sequence[str]],
+    denylist: Optional[Sequence[str]],
 ) -> Sequence[Tuple[str, ServiceNamespaceConfig]]:
```
stug, line 1023: oh good call making this a pure function

src/tests/configure_synapse_test.py
```diff
@@ -1748,13 +1748,20 @@ def test_get_allowed_namespaces(
...
-        allowed_namespaces = configure_synapse.get_allowed_namespaces(
-            ['allowed_service'],
-        )
-
-        assert allowed_namespaces == [allowed_service]
+    assert configure_synapse.get_allowed_namespaces(
+        [allowed_service, disallowed_service],
+        ['allowed_service'],
+        None,
+    ) == [allowed_service]
```
stug, line 1755: nit: can you maybe call these as kwargs for readability?

stug (approved):
lg2m!  And yeah I agree the denylist is a better interface for this probably.

vkhromov (approved):

PR #108, opened 2021-11-05
Use re.search to match allow/deny lists

keymone:
Changes we want to ignore are usually related to added/removed namespaces, this means we'd need to update allow/deny lists all the time for these filters to take effect. Using regular expressions we can address this.

`allowlist` and `denylist` are interpreted as regular expressions (special characters need to be escaped to have literal meaning) and are meant to be prefix-matchers (`^` is always prepended).


src/synapse_tools/configure_synapse.py
```diff
@@ -1016,6 +1017,13 @@ def get_envoy_migration_config(path: str) -> EnvoyMigrationConfig:
         return yaml.load(f, Loader=yaml.CSafeLoader)
 
 
+def re_search_in(coll: Sequence[str], item: str) -> bool:
+    for test in coll:
+        if re.search(f"^{test}", item):
```
vkhromov, line 1022: nit:
It's better to use `re.match("re",` instead of `re.search("^re",`.
And probably better to replace the whole function body with something like
```
    return re.match("(?:" + ")|(?:".join(coll) + ")", item)
```

vkhromov (approved):

vkhromov (approved):

mattmb (approved):

PR #109, opened 2021-11-10
do the right thing when listen_to_haproxy flag is set to false

keymone:


TheNavigat:
Can you add a test that involves setting `listen_with_haproxy` and verifying the generated configuration, please?

TheNavigat (approved):

PR #117, opened 2022-01-18
U/buffer0x7cd/fix xenial dockerfile

Buffer0x7cd:
use yelp image for xenial


dockerfiles/xenial/Dockerfile
```diff
@@ -1,20 +1,18 @@
-FROM ubuntu:xenial
+FROM docker-dev.yelpcorp.com/xenial_pkgbuild
```
danielpops, line 1: is this intended to be a temporary or permanent change?

Buffer0x7cd: This is temporary change to fix the build, we will move it to https://github.com/Yelp/clusterman/blob/fc4e52ab01b3b4134f5913af49bcd9671ad49651/Makefile#L78 this format once we have fixed other builds

nemacysts: deadsnakes has dropped all xenial debs, so somewhat permanent (unless we host our own xenial debs)

nemacysts (approved):

dockerfiles/xenial/Dockerfile
TheNavigat: I would assume permanent (or temporary until we deprecate Xenial), since Xenial is dropped from GHA (and from public usage)

danielpops (approved):

